Issue #19, opened 2017-11-04
Implement seek() methods in TextFileReader


davidpablocohn:

MignonBelongie:
I think I should do this one before #20.

davidpablocohn:
Makes sense to me - I put in a stub with the tempate in
loggers/readers/reader.py:StorageReader(), but that's not necessarily how
it needs to go. Thanks!

PR #48, opened 2018-01-02
Issue #19: implemented seek() and read_range() in TextFileReader

MignonBelongie:
There are still a few TODO's in the code (and a question), but it's mostly working, so I think it's ready for feedback.

### DIFF
--- a/logger/readers/text_file_reader.py
+++ b/logger/readers/text_file_reader.py
@@ -73,6 +73,10 @@ class TextFileReader(StorageReader):
     # The file we're currently using
     self.current_file = None
 
+    self.pos = 0
+    self.start_pos = {}
+    self.end_pos = {}
+
   ############################
   def _get_next_file(self):
     """Internal - Open and assign the next unused file to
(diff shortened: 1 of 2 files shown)


logger/readers/text_file_reader.py
```diff
@@ -147,3 +158,150 @@ def read(self):
       logging.debug('TextFileReader - tail/refresh specified, so sleeping '
                     '%f seconds before trying again', self.retry_interval)
       time.sleep(self.retry_interval)
+
+  ############################
+  # Current behavior is to just go to the end if we run out of records,
+  # as io.IOBase.seek() does.
+  # QUESTION: To really behave like seek(), we'd have to keep track of self.pos
+  # beyond the end of the file, e.g. seek(100, 'start') would always return
+  # 100, even if there are < 100 records. Is this what we want?
```
davidpablocohn: Going to the end of current input makes the most intuitive sense to me. Not sure I understand the semantics of seeking beyond the end...(though it sounds like I should).

logger/readers/text_file_reader.py
```diff
@@ -147,3 +158,150 @@ def read(self):
...
+      self.current_file.seek(state['current_file_pos'])
+    else:
+      self.current_file = None
+    self.pos = state['pos']
+
+  ############################
+  # Behavior is intended to mimic file seek() behavior but with
+  # respect to records: 'offset' means number of records, and origin
+  # is either 'start', 'current' or 'end'.
+  def seek(self, offset=0, origin='current'):
```
davidpablocohn: I realize you've made this more powerful than I'd thought about. For some reason, I'd not considered seeking beyond the *current* file, but of course it makes perfect sense to do so: when the TextFileReader is initialized with multiple files, the user shouldn't be expected to know *which* file is the current one. So nicely done on that front!

logger/readers/text_file_reader.py
```diff
@@ -147,3 +158,150 @@ def read(self):
...
+        pos = 0
+        for filename in file_list:
+          if filename in self.end_pos:
+            pos = self.end_pos[filename]
+          else:
+            self.start_pos[filename] = pos
+
+            # TODO: this can be made faster, if needed
+            with open(filename) as f:
+              for n, _ in enumerate(f, 1):
```
davidpablocohn: Very cool - I'd never seen this trick before.

PR #52, opened 2018-02-19
Implemented seek_time() and read_time_range() for LogfileReader.

MignonBelongie:
I finally had some time to work on this. It's not perfect, but at least this time it matches the definitions and descriptions of seek_time() and read_time_range() in TimestampedReader.

Caveats:
1) I didn't address the issue of out-of-order records. I think the hardest part of that is just deciding what the desired behavior is. The current behavior is probably adequate for many use cases as long as the records are approximately sorted.
2) It doesn't do things like skip over entire files based on timestamps, or binary search within a file. In order to do that, I think it would be necessary for LogfileReader to stop using TextFileReader and instead do its own file handling, because TextFileReader doesn't know about time, and even if it did, skipping over a file (without counting the records in it) would break its own model. Even as is, LogfileReader is making the questionable assumption that TextFileReader is iterating through files in chronological order. I don't think this change would be that hard to implement, but I think some discussion is in order, first.

### DIFF
--- a/logger/readers/logfile_reader.py
+++ b/logger/readers/logfile_reader.py
@@ -22,6 +22,7 @@ class LogfileReader(TimestampedReader):
   ############################
   def __init__(self, filebase=None, tail=False, refresh_file_spec=False,
                retry_interval=0.1, interval=0, use_timestamps=False,
+               time_format=timestamp.TIME_FORMAT,
                date_format=timestamp.DATE_FORMAT):
     """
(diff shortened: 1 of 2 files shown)


davidpablocohn:
Hurrah - thank you! Looks great to me.

I'm in transit today, but will take it out for a spin tomorrow. I started
working on DatabaseWriter and DatabaseReader (another TimestampedReader),
and think I'm pretty close.

PR #109, opened 2019-05-10
specify custom separator

webbpinner:
Add optional custom separator between timestamp and data string


webbpinner:
Easy add-on

davidpablocohn:
Thanks! Precisely what I was going to suggest. But would you be willing to change the keyword from 'separator' to 'sep' to be consistent with convention used by the python split() method and the [logger/transforms/prefix_transform.py](logger/transforms/prefix_transform.py)?

webbpinner:
ok... change completed

davidpablocohn:
Thanks for this! Merged.

PR #111, opened 2019-05-11
Introduction of a polled serial reader

webbpinner:
Polled serial reader extends the serial reader to include options for defining an initialzation string to be sent to the host at startup, a string to be sent to the host prior to reading new serial data (some sensors require a command to be sent to them prior to transmitting data back) and a string to send to the host when the reader is destroyed.

This is completely untested


davidpablocohn:
Julian - can you give Webb's two fixes (thank you, Webb!) a try on your
serial readers and let us know whether they meet your needs?

PR #171, opened 2019-10-24
Issue 166

MignonBelongie:
Here's an incomplete but working implementation of a TimelineWidget, ready for feedback.

Comments:

- We should probably have some kind of build process. (I'm willing to work on that.)
- I made my code conform to https://standardjs.com (as specified in CONTRIBUTING.md).
- The only standardjs errors remaining in d3_widgets.js are "'d3' is not defined". Ideally, I would import D3 as a module instead of including it as a script, which would solve this. Unfortunately, D3 is not released in ES6 module format, because they expect you to build it as part of your build process. (See first comment.)


davidpablocohn:
This looks great as a first/second/third pass! I'm thinking it probably
makes sense to go ahead and check this in as it is, and let you go ahead
and refine things as time allows. I'm wondering if it makes sense to open a
set of sub-feature requests around the widget as things go forward?

My JS skills are minimal, but I'm wondering about the difference between
including code via the   "<script src="whatever">" directive and using the
"import from whatever" JS statement. I didn't even know about the latter.
Is there something you can me to that I can use to read up on it?

(I notice that the graphs have a jitter that Valerie wrestled with a bunch.
Every once in a while, things seem to jump back just a touch. I recall that
Valerie figured out how to suppress it, but I don't recall how.)

MignonBelongie:
I didn't think it was ready to merge, but I suppose it's OK since no one has to actually use it yet. Opening new bug reports/feature requests sounds good.

Here's a good reference for ES6 modules: https://exploringjs.com/es6/ch_modules.html
I implemented d3_widgets.js as a module, so that's one example you could look at.

About the jitter: I haven't yet studied Valerie's code. I decided to push before addressing the jitter, because it looks like a non-trivial problem, and for me, at least, it's mostly transient, i.e., once it finally stops jittering, it stays smooth. However, I certainly would like to fix it!

davidpablocohn:
Sorry - the merge was premature on my part, but I decided that, since it
was all auxiliary code, there was no harm in leaving it merged. Also I
should have merged it to the dev branch. I need to put something into the
"contributing" docs about working from the dev branch. My model is that we
primarily work on dev, and merge dev->master for numbered version releases.

Many thanks again for working on this!

PR #173, opened 2019-10-29
D3 based DialWidget

MignonBelongie:
ready for feedback

### DIFF
--- a/display/html/d3_demo.html
+++ b/display/html/d3_demo.html
@@ -44,6 +44,49 @@
       g.axis.y path.domain {
         stroke: none;
       }
+      svg.gauge {
+        margin: 0 auto;
+        display: block;
+      }
+      svg.gauge circle.outer {
+        fill: #ccc;
+        stroke: #666;
+      }
+      svg.gauge circle.inner {
+        fill: #eee;
+        stroke: #aaa;
+      }
+      svg.gauge path.arc {
+        stroke: #666;
+        stroke-width: 2px;
+      }
+      svg.gauge path.wedge {
+        fill: #ccc;
+      }
+      svg.gauge circle.pin {
+        fill: #222;
+      }
+      svg.gauge path.pointer {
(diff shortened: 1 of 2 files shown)


davidpablocohn:
Sorry for the delay on this! Codewise, via inspection and trial, it all
looks good, and you keep teaching me more about JS - I had to look up the
"..." operator, and wish I'd known about that ages ago.

Dial looks great, and almost ready to go. I could imagine a few more
options prior to integration, the main one being an option of whether to
show the numeric value(s) in the center of the gauge, and perhaps whether
to also include the name along with the numeric value. Other than that, it
looks fabulous.

On the timeline, I think the main features I'd like to see are a few more
widget options:
 - min, max  - optional limits on max min display values
 - minRange - minimum difference between max and min

Again, this all looks great - many thanks for taking it on!

MignonBelongie:
I added options 'showNumericValues' and 'showDescriptions'. I'm a little concerned that if both are false, there is no way to see the descriptions. I don't know if you noticed, but there were tooltips on the numeric values (and they're still there if showNumericValues is true and showDescriptions is false). I tried putting tooltips on the pointers, but I just couldn't get it to work right.
One thought is to have a "more info" button that would show the list of fields, and maybe even allow deselecting fields. New feature request, maybe?
Finally, I set the defaults to match the original behavior, but feel free to change them (or any of the other defaults I pulled out of a hat).

davidpablocohn:
I'd missed the tool tips - thanks!

In terms of omitting descriptions, I think lots of times folks will want
the descriptions to be external to the widget, e.g. as a separate text
widget. So I think that's actually a feature rather than a bug.

I think it looks ready to merge with dev as a first pass, if you're happy
with it.  Let me know?

Thanks, too, for opening the "additional features" issue report.

MignonBelongie:
If you're OK with the defaults, go ahead and merge.

PR #179, opened 2019-11-26
Make D3DialWidget center itself in its containing div.

MignonBelongie:
This works for me now with a fixed size div. I'm sure there's a simpler way to do this, but hopefully this will work for you.


davidpablocohn:
Perfect - thanks!

PR #181, opened 2019-12-07
added more options for D3 TimelineWidget

MignonBelongie:
I added all the options you asked for (plus a few related ones). I also edited one of the TimelineWidgets in nbp_dashboard_d3.html. You can play around with the settings, and then make the same changes in all the TimelineWidgets. However, for the font size and background color settings, it probably makes more sense to do it with CSS, if they're all going to be the same. You might not actually need to set numXTicks, because the reason there were so many ticks before is that I had hardcoded it to make tick marks every 15 seconds - sorry about that!

### DIFF
--- a/display/html/d3_demo.html
+++ b/display/html/d3_demo.html
@@ -179,6 +179,14 @@
           //max: 20,
           //minRange: 5,
           //showLegend: false,
+          //backgroundColor: 'lightblue',
+          //xTickFontSize: 8,
+          //xTickFormat: '%H:%M:%S',
+          //numXTicks: 5,
+          //yTickFontSize: 8,
+          //yTickFormat: '.2e',
+          //numYTicks: 5,
+          //yUnitsFontSize: 12,
           //maxPoints: 30,
         }
       ));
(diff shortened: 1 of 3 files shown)


davidpablocohn:
These look great, thank you! I've patched, based on your example, into http://165.227.31.102:8000/display/nbp_dashboard_d3.html, and am ready to merge into dev when you give me the thumbs up.

The one thing that doesn't seem to be working the way I'd expect is the new option yUnitsFontSize, which I would expect to change the font size of the text like "Sea Temp C" along the left side of the graph. yTickFontSize appears to properly control the size of the text on the ticks, but yUnitsFontSize doesn't seem to have any effect. For example, on the above page, I've got yUnitsFontSize set to 10 for the "Air Temp C" line, and set to 30 for the "Sea Temp C" line. Any idea what I'm doing wrong?

MignonBelongie:
It looks like the problem is that yUnitsFontSize was already included in the HTML config, so that later setting is overriding the one you added.

davidpablocohn:
Oooh - that's embarrassing. Thanks for catching that!

PR #191, opened 2020-01-14
Add ability to truncate log files by hour

webbpinner:
Some users may want hourly files instead of daily.  This will be true for larger datasets.  These changes allow users to easily declare if they want hourly files.

### DIFF
--- a/logger/writers/logfile_writer.py
+++ b/logger/writers/logfile_writer.py
@@ -14,20 +14,25 @@ class LogfileWriter(Writer):
   """Write to the specified file. If filename is empty, write to stdout."""
   def __init__(self, filebase=None, flush=True,
                time_format=timestamp.TIME_FORMAT,
-               date_format=timestamp.DATE_FORMAT):
+               date_format=timestamp.DATE_FORMAT,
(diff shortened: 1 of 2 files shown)


davidpablocohn:
If I'm reading the commits correctly, #189 and #191 are compatible in concept, but I don't think Git will let me merge both, since they modify overlapping bits of code (specifically the function argument definition line):

```
  def __init__(self, filebase=None, flush=True,
               time_format=timestamp.TIME_FORMAT,
               date_format=timestamp.DATE_FORMAT,
               rollover_hourly=False):
```
and
```
  def __init__(self, filebase=None, flush=True,
               time_format=timestamp.TIME_FORMAT,
               date_format=timestamp.DATE_FORMAT,
               suffix=''):
```
If I am reading it right, could you merge the two commits, and then I'll squash/merge the combined pull request?

webbpinner:
No worries... just commit one of them and I'll re-fork and make the necessary modifications to the new master.

webbpinner:
combined the suffix and hourly_rollover into a single branch/pull request.

PR #193, opened 2020-01-16
Initial import of influxDB_writer

webbpinner:
Initial import of influxDB_writer, settings file template and associated changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ django_gui/migrations
 
 django_gui/settings.py
 database/settings.py
+logger/settings.py
 hapi/settings.py
 display/js/widgets/settings.js
 static/
--- a/logger/listener/listen.py
+++ b/logger/listener/listen.py
@@ -87,6 +87,7 @@ from logger.writers.udp_writer import UDPWriter
 from logger.writers.redis_writer import RedisWriter
 from logger.writers.text_file_writer import TextFileWriter
 from logger.writers.logfile_writer import LogfileWriter
+from logger.writers.influxdb_writer import InfluxDBWriter
 from logger.writers.database_writer import DatabaseWriter
 from logger.writers.record_screen_writer import RecordScreenWriter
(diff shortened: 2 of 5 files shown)


davidpablocohn:
Overall, looks great, with a few minor tweaks:

1. It looks like the block comments in influxDB_writer refer to an earlier template ("""Note: Assume record begins with a timestamp string."""), while the code itself assumes a dict or DASRecord.

2. I'd prefer that the write() method check its input format as being dict (or DASRecord) so that it complains rather than throwing a TypeError.

3. If it is a dict or DASRecord, it should have its own timestamp already; I'd prefer we use that rather than current system time. This is important for things like the InterpolationTransform, which may emit records with timestamps that are several seconds older than the current system time (because they are averaging around a window).

4. I'm trying to convert all transforms and writers to be able to take lists of records in addition to individual records. For writers, this is an easy fix, as in text_file_writer.py's write() method:
```
    # If we've got a list, hope it's a list of records. Recurse,
    # calling write() on each of the list elements in order.
    if type(record) is list:
      for single_record in record:
        self.write(single_record)
      return
```

webbpinner:
Thanks for taking a look... I'll try to address these issues.

number 1 -> copy/paste error from logfile_writer, will update
number 2 -> I knew this was going to be an issue... would like to discuss how best to address it.
number 3 -> That's easy enough to address.
number 4 -> InfluxDB can take either a dict or array of dict.

webbpinner:
Ok... please take a look at the lastest commit on this branch.  Should address issues 1, 3, and 4.

davidpablocohn:
Thanks! These look good. My practice on #2 is to have the write() method
check each record and log a warning if it's not in an expected format. See,
e.g. logger/writers/cached_data_writer.py. As per the comments in that
class write() method, I'm hoping to phase out the [(ts, value), (ts,
value),...] twist on on dict field values, so I wouldn't bother checking
for and handling those.

Shall I go ahead and accept the pull now, or wait for you to field #2?

PR #213, opened 2020-04-27
Issue 53

pandeliskirpoglou:
MQTT readers and tests

### DIFF
--- a/logger/readers/mqtt_reader.py
+++ b/logger/readers/mqtt_reader.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+
+
+import json
+import logging
+import socket
+import sys
+
+from os.path import dirname, realpath; sys.path.append(dirname(dirname(dirname(realpath(__file__)))))
+
+from logger.utils.formats import Text
+from logger.readers.reader import Reader
+
+try:
+  import paho.mqtt.client as mqtt # import the client | $ pip installing paho-mqtt is necessary
+  PAHO_ENABLED = True
+except ModuleNotFoundError:
+  PAHO_ENABLED = False
+
+################################################################################
+class MQTTReader(Reader):
+  """
+   
+  """
+  def __init__(self, broker, channel):
+
+    super().__init__(output_format=Text)
+
+    if not PAHO_ENABLED:
+      raise ModuleNotFoundError('MQTTReader(): paho-mqtt is not installed. Please '
+                                'try "pip install paho-mqtt" prior to use.')
+    
+    self.broker = broker
+    self.channel = channel
+    
+    try:
+      self.paho = mqtt.Client('Instance1')
+
+      self.paho.connect(broker)
+      self.paho.subscribe(channel)
+
+    except mqtt.WebsocketConnectionError:
+      print('Unable to connect to broker at %s:%d',
+                    self.broker)
+
+  ############################
+def read(self):
+  while True:
+    try:
+      self.paho.loop_forever()
+      message = next(iter(self.paho.listen()))
+      logging.debug('Got message "%s"', message)
+      if message.get('type', None) == 'message':
+        data = message.get('data', None)
+        if data:
+          return data
+    except KeyboardInterrupt:
+      self.paho.disconnect()
+      exit(0)
+
+################################################################################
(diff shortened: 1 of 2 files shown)


logger/readers/test_mqtt_reader.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+
+
+from logger.readers.mqtt_reader import MQTTReader
```
davidpablocohn, line 4: The ``from logger.readers.mqtt_reader`` include should go after the sys.path.append() below so that the unit test can be run standalone, e.g.

```
python3 logger/readers/test_mqtt_reader.py
```

logger/readers/test_mqtt_reader.py
```diff
@@ -0,0 +1,72 @@
...
+broker_address = 'test_mqtt_reader'
+channel = 'path/of/devices'
+
+##############################################################################
+
+
+class TestMQTTReader(unittest.TestCase):
+
+    @unittest.skipUnless(PAHO_ENABLED, 'Paho MQTT not installed; tests of MQTT '
+                         'functionality will not be run.')
```
davidpablocohn, line 44: I guess I don't understand MQTT brokers - is there a need to actually instantiate the broker in the test and set it running in a separate thread, so that the reader (and future writers) can connect to it? Or is it instantiated ad hoc? Apologies - as I said, I don't particularly understand MQTT.

logger/readers/mqtt_reader.py
```diff
@@ -0,0 +1,61 @@
...
+try:
+  import paho.mqtt.client as mqtt # import the client | $ pip installing paho-mqtt is necessary
+  PAHO_ENABLED = True
+except ModuleNotFoundError:
+  PAHO_ENABLED = False
+
+################################################################################
+class MQTTReader(Reader):
+  """
+
```
davidpablocohn, line 23: Could you add a docstring here with explanation of parameters?

logger/readers/mqtt_reader.py
```diff
@@ -0,0 +1,61 @@
...
+    self.channel = channel
+    
+    try:
+      self.paho = mqtt.Client('Instance1')
+
+      self.paho.connect(broker)
+      self.paho.subscribe(channel)
+
+    except mqtt.WebsocketConnectionError:
+      print('Unable to connect to broker at %s:%d',
```
davidpablocohn, line 43: Let's use logging.error() instead of print()

logger/readers/mqtt_reader.py
pandeliskirpoglou: Okay. My bad totally forgot.

logger/readers/mqtt_reader.py
pandeliskirpoglou: I changed it last moment cause I was having some problems with the compilation but sure can do!

logger/readers/test_mqtt_reader.py
pandeliskirpoglou: Well, I had the same concern in that matter. I will look more into that.

davidpablocohn:
You might be able to take some inspiration from the code at
logger/readers/test_redis_reader.py, which checks if there's a redis server
up and running, and starts one if there isn't.

Thanks!

davidpablocohn:
I'm trying to manually start a broker that the reader/writer can connect to for the test and can't figure out how. (If you run the test with paho-mqtt running, but no broker running, it throws a socket error at

```
  File "/opt/openrvdas/logger/writers/mqtt_writer.py", line 45, in __init__
    self.paho.connect(broker)
```

Could you 

1) let me know how I can manually start a broker?
2) possibly put that in the docstring for the test?

Thanks!

pandeliskirpoglou:
Instructions on how to start an MQTT broker:

    1. First install the Mosquitto Broker :
	```
	sudo apt-get update
	sudo apt-get install mosquitto
	sudo apt-get install mosquitto-clients
	```
    2. The mosquitto service starts automatically when downloaded but use :
	```
	sudo service mosquitto start
	sudo service mosquitto stop
	```
	to start and stop the service.
	
    3. To test the install use:
	```
	netstat -at  
	```
	and you should see the MQTT broker which is the port 1883
	
    4. In order to manually subscribe to a client use : 
	```
	mosquitto_sub -t "example/topic"
	```
 	and publish a message by using
	```
	mosquitto_pub -m "published message" -t "certain/channel"
	```
    5. Mosquitto uses a configuration file "mosquitto.conf" which you can find in    /etc/mosquitto folder

PR #217, opened 2020-06-20
NMEA Checksum Transform

anshika-agarwal:
test_nmea_checksum_transform.py is the unittest script for nmea_checksum_transform.py

### DIFF
--- a/nmea_checksum_transform.py
+++ b/nmea_checksum_transform.py
@@ -1,0 +1,114 @@
+import logging
+import sys
+
+# For efficient checksum code
+from functools import reduce
+from operator import xor
+
+from logger.writers.writer import Writer
+
+############################
+
+def get_checksum_str(source):
+    """ Returns checksum_str, which is everything between the '$' and '*' in the source string """
+    
+    if ((source.find('$')==-1) or (source.find('*')==-1)):
+        return None
+    
+    start = source.index('$')+1
+    end = source.index('*')
+    checksum_str = source[start:end]
+    return checksum_str
+
+def get_checksum_value(source):
+    """ Returns checksum_value, which is the parsed checksum value (after '*') from the source string """
+    
+    if (source.find('*')==-1):
+        return None
+    
+    start = source.index('*')+1
+    checksum_value = source[start:]
+    return checksum_value
+
+def compute_checksum(source):
+    """Return hex checksum for source string."""
+    
+    return '%02X' % reduce(xor, (ord(c) for c in source))
+
+################################################################################
+class NMEAChecksumTransform:
+    """
+        NMEAChecksumTransform checks the integrity/completeness of a record by confirming whether or not the checksum matches. If the checksum matches, it returns the record, and otherwise it sends an error message.
+    """
+
+    DEFAULT_ERROR_MESSAGE = 'Bad checksum for record: '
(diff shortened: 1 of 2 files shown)


nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+                        '(type %s): %s', type(record), record)
+            return None
+        
+        message_str = get_checksum_str(record)
+        checksum_str = get_checksum_str(record)
+
+        checksum_value = get_checksum_value(record)
+        computed_checksum = compute_checksum(message_str)
+        
+        if (checksum_str == None):
```
davidpablocohn, line 83: Should be (checksum_value == None)? Because checksum_str is the source string that is to be checksummed, right?

nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+# For efficient checksum code
+from functools import reduce
+from operator import xor
+
+from logger.writers.writer import Writer
+
+############################
+
+def get_checksum_str(source):
```
davidpablocohn, line 12: I'd probably call this get_message_str() because it gets the message part of the string.

nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+        if not record:
+            return None
+    
+        if not type(record) is str:
+            logging.warning('NMEAChecksumTransform passed non-string record '
+                        '(type %s): %s', type(record), record)
+            return None
+        
+        message_str = get_checksum_str(record)
+        checksum_str = get_checksum_str(record)
```
davidpablocohn, line 78: This line/variable is redundant with message_str, right? The only place I see it used is ``if (checksum_str == None)``, below, where I think you actually want ``if (checksum_value == None)``.

nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def send_error_message(self, record, message=None):
+        """
+            Send error to error_writer if one exists, otherwise send it to stderr
+            
+            record - the record with the error
+            message - optional custom message. If None, then use self.error_message
+        """
+        
+        error_message = self.error_message + record
```
davidpablocohn, line 107: HEre I think you only want to send self.error_message if they've not given you a custom message. So maybe 
```
error_message = message or self.error_message
error_message += record
```

nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+        
+        error_message = message or self.error_message
+        error_message += record
+        
+        if self.error_writer:
+            self.error_writr.write(error_message)
+        else:
+            logging.warning(error_message)
+                
+        return None
```
davidpablocohn, line 114: Don't actually need a return value for send_error_message(), as nothing looks at/stores it.

nmea_checksum_transform.py
```diff
@@ -0,0 +1,114 @@
...
+# For efficient checksum code
+from functools import reduce
+from operator import xor
+
+from logger.writers.writer import Writer
+
+############################
+
+def get_message_str(source):
+    """ Returns checksum_str, which is everything between the '$' and '*' in the source string """
```
davidpablocohn, line 13: Should probably be message_str instead of checksum_str here? Same below, in 20:21.

Issue #90, opened 2019-01-09
GUI-based construction of YAML loggers/configurations


davidpablocohn:
Users should not have to hand-code YAML/JSON to build loggers and cruise configurations. It would seem to be fairly straightforward to build a web-based interface that allows you to select and combine components into loggers, modes and cruises, either using a graphic drag-and-drop interface or, at the very least, a series of pulldown menus.

PR #218, opened 2020-06-25
Added a JSON/YAML Creator utilizing dropdown menus. Made with React

mchoi39:
Created a React application to create JSON and YAML configuration for OpenRVDAS. Added functionalities such as copy button in order to facilitate copy/pasting the configurations


davidpablocohn:
Thank you! Now that I'm thinking about the directory structure more, I
think it makes sense to have this under a top level 'contrib' directory.
E.g.  openrvdas/contrib/utils/JSON_YAML_Creator. I'll go ahead and run the
pull, then, if you don't mind, I'll move it over to contrib on the dev
branch.

davidpablocohn:
Linking to #90 .

Initial feedback:
 * default port 3000 used by Grafana; is this a Node default or something you've chosen? If something you've chosen, can you change to a non-conflicting port?
 * Once a reader has been selected, the the kwargs section says "select a reader" rather than selecting keyword arguments
 * When "PolledSerialReader" is selected, kwargs show up as "hello" and "works"
 * Same keyword arg can be added multiple times - shouldn't be allowed, as it creates a syntax error ('keyword repeated') in python
 * It looks like package-lock.json is managed/changed by npm. If that's the case, it shouldn't be part of the repo.

webbpinner:
Regarding the port 3000 concern...
This UI might be better implemented as a packaged react app hosted under nginx, apache, etc instead of with npm's development web-server (which is what's trying to run on port 3000).  Packaging react apps for production is a well documented process.  Take a look at how the sealog clients (also react-based) are packaged and installed.

Regarding package-lock.json...
Agreed, this should be excluded from the repo and added to the .gitignore file.

PR #220, opened 2020-06-29
Issue 90

mchoi39:
- Default port changed to 8000. This can be changed from .env file

- KWargs section says "Select keyword argument" instead of "Select a reader"

- PolledSerialReader now displays the correct KWargs

- Keyword args does not allow repeats

- package-lock.json excluded


davidpablocohn:
Ah - no, 8000 is worse. Port 8000 is the default port for the whole
OpenRVDAS console. I'm going to recommend that you have a look at
https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports
and
select a port # that doesn't appear anywhere on that well-known ports table.

Also: when I did the pull, I moved the source over to
contrib/utils/JSON_YAML_Creator on the dev branch. It looks like you've
moved it up one level, getting rid of the 'utils' subdir. Could you keep it
under 'utils'?

Thanks,
 -p

mchoi39:
How can I fix these conflicts with branch? All I did was move the folder one level down to contrib/utils/JSON_YAML_CREATOR 

I have tried following this to fix, but did not work:

git checkout dev
git pull
git checkout issue_90
git merge dev
git add .
git commit -m "msg"
git push

PR #222, opened 2020-07-03
CORIOLIX Writer Stub

webbpinner:
Migrated PostgreSQL naming to CORIOLIX as it is different then a PostgreSQL port of the MySQL database connector and writer.

Updated  setup.sh and update_cruise_definition.py  script to match. 
Adding Coriolix-specific settings.py.dist file.

None of this has been tested or even run.  The connector was ported from the PostgreSQL Connector.

### DIFF
--- a/local/rcrv/settings.py.dist
+++ b/local/rcrv/settings.py.dist
@@ -1,0 +1,42 @@
+"""
+Settings for coriolix operations.
+
+THIS FILE MUST BE COPIED OVER TO local/rcrv/settings.py to be
+operational.
+
+See below for additional coriolix-specific install requirements
+
+"""
+import logging
+import sys
+
+sys.path.append('.')
+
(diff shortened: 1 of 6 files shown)


local/rcrv/settings.py.dist
```diff
@@ -0,0 +1,42 @@
...
+"""
+import logging
+import sys
+
+sys.path.append('.')
+
+CORIOLIX_DATABASE = 'agg_openrvdas'
+CORIOLIX_DATABASE_HOST = 'localhost'
+CORIOLIX_DATABASE_USER = 'openrvdas_user'
+CORIOLIX_DATABASE_PASSWORD = 'coriolix_d@t@'
```
davidpablocohn, line 18: Do we want to excise this from the file in the repo?

webbpinner: Probably should generalize it... good call.

Issue #216, opened 2020-06-18
DeltaTransform


davidpablocohn:
We want to detect rapid changes in values. A couple of use cases from Tony D'aoust:

```
A section of .322 pulling a nice steady load up to 5,000 lbs is not a concern if the load is applied 
gradually over say 3 seconds of time.   If that same load is applied in a third of a second the wire 
lays do not have time to seat and get out of each other’s way and the section is at risk for more 
damage - particularly of course where it goes over blocks.

  I contend that it is more damaging to snap load the wire in micro seconds of loading time to 
loads well below the UNOLS 5,000 pound limit than it is to load the wire to that 5,000 pound limit
 but to do so in a controled fashion over several seconds of application.

  Does this make intuitive sense ?

  I’m not sure how simple the code would be to do this, but I feel strongly that this piece of 
information is missing in our operational assesments and wire documentation. 

     The script would be something that looks at say the last 10 seconds and rolls over including 
say 3 seconds with every new 10 second grouping.   It would just look at tension with respect to 
how fast that tension comes onto the wire.

  At 20 hz, the LCI90i could tell us for instance if at 1/20th of a second the CTD weighed 750 
pounds then at 2/20th weighed -350 pounds, then at 3/20th 150 lbs and then at 4/20th of a 
second the CTD now weighed 2,750 pounds in the water.   This would tell us quite dramatically
 just how violent the action of the package was.  One could say for instanc if one were inclined,
 "hey - this sucks - Abort Abort!” 

 Does this makes sense?
```

My first thoughts on implementation:

```
That makes excellent sense. I'll have to think of whether there's any way to piggyback onto
 existing transforms, but it would be completely straightforward to code up a new version of 
something like the QCFilterTransform (which checks for static values out of bounds). Say, call it a 
QCDeltaFilterTransform, and have it filter for sequences where values change by more than X in
 time T. It would be fairly generally useful, I think - could use it to flag when *any* type of sensor
 registers too big a change too quickly. E.g to detect whether a gyro might be failing (flipping 180
 degrees). Sample command line invocation would be something like: 

# Flag change of or trawl winch tension by more than 2000 lbs within 0.3 seconds
listener.py --serial port=/dev/tty_twnc \
      --transform_timestamp \
      --transform_parse_nmea \
      --transform_qc_delta_filter TWNCTension:2000:0.3

Could also do something a bit more sophisticated and specialized as well for winch tensions. E.g.
[...]

PR #235, opened 2020-07-28
Issue #216: DeltaTransform

anshika-agarwal:


delta_transform.py
```diff
@@ -0,0 +1,60 @@
...
+      return ('Record passed to DeltaTransform was neither a dict nor a '
+              'DASRecord. Type was %s: %s' % (type(record), str(record)[:80])) 
+    
+    for key in current_value:
+      delta_value = current_value[key] - last_value[key][1]
+      delta_time = timestamp - last_value[key][0]
+      if self.rate is True:
+        # return delta_value[key] / delta_time
+      else:
+        # return delta_value
```
anshika-agarwal, line 59: If a record has multiple fields that we would want to return deltas for, would it better to store them in something like results[] and just return that or call transform() recursively?

davidpablocohn:
What I'm recommending is isolation of logic so that the transform has two
functional pieces:

   1. Do what is necessary to get the data into a pair of variables for a
   single record: fields, timestamp
   This is where we'd see if it's a DASRecord and set fields =
   record.fields and timestamp = record.timestamp, or if it's a dict, set
   fields = record.get('fields', None). If it's a list, then we get to punt
   and recurse, calling ourselves on each element of the list, appending the
   results and returning right there. If we do it in order, we'll get the
   deltas between the successive records.

   2. Compute and return the delta of that record from the previous record

I guess we hadn't discussed what format the return should be. I think it
makes the most sense to return a DASRecord, where the timestamp is the ts
of the record being currently processed and the fields are the deltas. In
all generality, we should probably use different field names, since the
value no longer represents "heading" but rather "heading_delta". But let's
keep that for a future version.

But given the timestamp and the deltas in a dict called delta_values, you
can create/return a DASRecord via

  return DASRecord(timestamp=timestamp, fields=delta_values)

Does that make sense?

anshika-agarwal:
New comments resolved.

PR #243, opened 2020-08-30
Delta Transform and Unittest Script (Issue #216)

anshika-agarwal:
This pull request contains both files (delta_transform.py and test_delta_transform.py).


delta_transform.py
```diff
@@ -0,0 +1,72 @@
+import logging
```
davidpablocohn, line 1: It looks like the pull request includes both this delta_transform.py in the top-level directory (which should be removed) and a copy in logger/transforms. I'll put my comments in the copy that's in logger transforms. This copy should be deleted via 'git rm ./delta_transform.py'

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+from logger.utils.das_record import DASRecord
+from logger.utils import timestamp
+from logger.transforms.transform import Transform
+
+################################################################################
+
+class DeltaTransform:
+
+  # Dict of {field_name: (previous_timestamp, previous_value)} pairs
+  last_value_dict = {}
```
davidpablocohn, line 14: The way this is written, last_value_dict is a class variable - all instances will share the values there, which can lead to undesirable side effects. Instead, there should be one dict for each instance, so the values that one instance sees doesn't affect the values another sees. See below, in __init__()

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+
+  def __init__(self, rate=False, field_type=None):
+  """
+    rate — whether to return the rate (delta/second) for some or all fields, or just return the delta
+    field_type — if not None, should be a dict specifying special field types, if any
+  """
+      
+    self.rate = rate
+    self.field_type = field_type
+
```
davidpablocohn, line 27: self.last_value_dict = {}  # should be here

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+      return None
+    
+    fields = {}
+
+    if type(record) is DASRecord:
+      fields = record.fields
+      timestamp = record.timestamp 
+    
+    elif type(record) is dict:
+      fields = record['fields']
```
davidpablocohn, line 41: I'd prefer the version that you've got commented out below. "fields = record['fields']" will throw a KeyError and crash the whole logger if it gets a record without  a 'fields' entry. If you use "fields = record.get('fields', None)" then you can check
```
if fields is None:
  logging.info('Received record without "fields": %s', record)
  return None
```
and likewise with timestamp, to ignore unusable records.

(That check is down in line 58, but I'd change it to check for fields and timestamp separately to aid the user in knowing which is missing, and would also display the faulty record, as above.)

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+      #timestamp = record.get('timestamp', None)
+      
+    elif type(record) is list:
+      results = []
+      for single_record in record:
+       results.append(self.transform(single_record))
+      return results
+    
+    else:
+      return ('Record passed to DeltaTransform was neither a dict nor a '
```
davidpablocohn, line 55: Again, here you should do a logging.info() of the error message, and return None. In current form, it will drop the error message into the record stream, and downstream transforms/writers won't know what to do with it.

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+              'DASRecord. Type was %s: %s' % (type(record), str(record)[:80])) 
+    
+    if fields is None or timestamp is None:
```
davidpablocohn, line 58: As mentioned in the comment above, I'd do these checks separately and send the record to stderr to help user figure out what was wrong:
```
if fields is None:
  logging.info('Received record without "fields": %s', record)
  return None
if timestamp is None:
  logging.info('Received record without timestamp: %s', record)
  return None
```
And again - transforms should return None if there's a problem with the record.

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+      return ('Record passed to DeltaTransform either does not have a field or '
+              'a timestamp')
+    
+    delta_values = {}
+    rate_values = {}
+    
+    for key, value in fields.items():
+
```
davidpablocohn, line 66: delete blank line here for style

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+    if fields is None or timestamp is None:
+      return ('Record passed to DeltaTransform either does not have a field or '
+              'a timestamp')
+    
+    delta_values = {}
+    rate_values = {}
+    
+    for key, value in fields.items():
+        
+      if key in DeltaTransform.last_value_dict:
```
davidpablocohn, line 67: Assuming you move last_value_dict to being an instance variable, this should become
```
if key in self.last_value_dict:
```

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+            if self.field_type[key] == 'polar':
+              delta_values[key] = DeltaTransform.polar_diff(self, last_value, value)
+          else:
+            delta_values[key] = value - last_value
+        else:
+          return ('field_type passed to DeltaTransform is neither None nor a dict')
+
+        rate_values[key] = delta_values[key] / (timestamp - last_timestamp)
+
+        DeltaTransform.last_value_dict[key] = (timestamp, value)
```
davidpablocohn, line 84: Should become ``self.last_value_dict[key]`` along with other occurrences below.

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+
+    if self.rate:
+      return DASRecord(timestamp=timestamp, fields=rate_values)
+    elif type(self.rate) is list:
+      results = {}
+      for field in self.rate:
+        results[field] = rate_values[field]
+      return DASRecord(timestamp=timestamp, fields=results)
+    else:
+
```
davidpablocohn, line 98: Remove blank.

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+from logger.transforms.transform import Transform
+
+################################################################################
+
+class DeltaTransform:
+
+  # Dict of {field_name: (previous_timestamp, previous_value)} pairs
+  last_value_dict = {}
+
+  def polar_diff(self, last_value, value):
```
davidpablocohn, line 16: If you move this outside of the DeltaTransform, you can get rid of all the 'self' stuff, and just have it be ``polar_diff(self, last_value, value)``

logger/transforms/delta_transform.py
```diff
@@ -0,0 +1,99 @@
...
+      if key in DeltaTransform.last_value_dict:
+          
+        last_timestamp, last_value = DeltaTransform.last_value_dict.get(key, (None, None))
+            
+        if self.field_type is None:
+          delta_values[key] = value - last_value
+        elif type(self.field_type) is dict:
+          if key in self.field_type:
+            if self.field_type[key] == 'polar':
+              delta_values[key] = DeltaTransform.polar_diff(self, last_value, value)
```
davidpablocohn, line 76: As per comment on line 16, if you move polar_diff outside of DeltaTransform, this becomes just
```
delta_values[key] = polar_diff(last_value, value)
```

logger/transforms/test_delta_transform.py
```diff
@@ -0,0 +1,41 @@
...
+
+from os.path import dirname, realpath; sys.path.append(dirname(dirname(dirname(realpath(__file__)))))
+
+from logger.utils.das_record import DASRecord
+from logger.utils import timestamp
+from logger.transforms.delta_transform import DeltaTransform
+
+class TestDeltaTransform(unittest.TestCase):
+
+    def test_default(self):
```
davidpablocohn, line 15: I'd break this into two different tests: one creates a DeltaTransform for polar values, and one for non-polar values. Similarly for rate=True

To test properly, each t should be called multiple times. Recall that the first time you call a DeltaTransform there will have been no previous record to compute the delta with, so it should probably return None. The next time you call it, you should get the delta between the previous record and the current one:
```
t = DeltaTransform()
# First record, there's nothing to "delta" with
self.assertEqual(t.transform({'timestamp': 1, 'fields': {'variable': 15}}, None)

# Second record, at timestamp 2, has a delta of +5
self.assertEqual(
  t.transform({'timestamp': 2, 'fields': {'variable': 20}}),
              {'timestamp': 2, 'fields': {'variable': 5}})
# Third record, at timestamp 10, has a delta of -30
self.assertEqual(
  t.transform({'timestamp': 10, 'fields': {'variable': -10}}),
              {'timestamp': 10, 'fields': {'variable': -30}})
```
And likewise for the polar one and for the rate one. Note how I used two different time intervals, so that we can catch if rate is being inadvertently applied.

logger/transforms/test_delta_transform.py
```diff
@@ -0,0 +1,41 @@
...
+
+import logging
+import sys
+import unittest
+
+from os.path import dirname, realpath; sys.path.append(dirname(dirname(dirname(realpath(__file__)))))
+
+from logger.utils.das_record import DASRecord
+from logger.utils import timestamp
+from logger.transforms.delta_transform import DeltaTransform
```
davidpablocohn, line 11: If you move polar_transform up to the top level (as mentioned in the previous file comments), this would become
```
from logger.transforms.delta_transform import polar_diff, DeltaTransform
```

server/test_cached_data_server.py
```diff
@@ -227,6 +227,110 @@ async def run_test():
     asyncio.new_event_loop().run_until_complete(run_test())
     time.sleep(1)
 
+  ############################
```
davidpablocohn, line 230: It looks like you've got two copies of 'test_wildcard' in this file?

davidpablocohn (approved):

PR #270, opened 2021-06-18
Serial Writer

ghostcat17:
Just an implementation of the serial writer module. I got rid of a couple of unnecessary parameters(e.g. max_bytes, eol, etc.), added another exception if the encoding type was not provided, added a quick summary of the functionality of the class, and also implemented a write() method.


logger/writers/serial_writer.py
```diff
@@ -0,0 +1,76 @@
...
+            sys.exit(1)
+
+       
+        self.encoding = encoding
+        self.encoding_errors = encoding_errors
+
+
+    ############################
+    def write(self, record):
+        try:
```
davidpablocohn, line 65: Want to guard upfront against empty records (which often show up if, for example, downstream of a filter transform):

    if not record:
        return

Otherwise record.encode() will barf.

(Cleanest, I suppose to put before the 'try')

logger/writers/serial_writer.py
```diff
@@ -0,0 +1,76 @@
...
+            written = self.serial.write(record.encode(self.encoding))
+            if not written:
+                return None
+            return written
```
davidpablocohn, line 69: The generic Writer() API doesn't define any return value. Thinking it would be better to skip the 'return's and log a warning if not all bytes are written. But also provide a 'quiet' flag in the constructor to allow silencing those warnings.

davidpablocohn (commented):
A few comments below, but overall, nicely done! Let's hold off merging until there's a unittest, and we can pull them together.

Thanks!

davidpablocohn (approved):

PR #292, opened 2021-11-02
File writer - interval

webbpinner:
Adds ability to truncate files every x hours or x minutes.  Before committing please ponder any negative performance impacts cause by the more complex processing of the file name... this is working for my test case but may be a problem for higher data rate streams.

### DIFF
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -148,7 +148,6 @@ class FileWriter(Writer):
             timestamp_raw = datetime.now(time_zone)
             timestamp_proc = timestamp_raw.replace(hour=self.split_interval[0] * math.floor(timestamp_raw.hour/self.split_interval[0]), minute=0, second=0)
 
-            print(timestamp_proc.timestamp())
             return time_str(timestamp=timestamp_proc.timestamp(), time_zone=self.time_zone,
                             time_format=self.time_format)
 
@@ -157,7 +156,6 @@ class FileWriter(Writer):
             timestamp_raw = datetime.now(self.time_zone)
             timestamp_proc = timestamp_raw.replace(minute=self.split_interval[0] * math.floor(timestamp_raw.minute/self.split_interval[0]), second=0)
 
-            print(timestamp_proc.timestamp())
             return time_str(timestamp=timestamp_proc.timestamp(), time_zone=self.time_zone,
                             time_format=self.time_format)
 
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -3,7 +3,6 @@
 import os.path
 import sys
 import math
-import logging
 
 from datetime import datetime, timezone
 
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -4,7 +4,7 @@ import os.path
 import sys
 import math
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 from os.path import dirname, realpath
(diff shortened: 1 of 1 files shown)


logger/writers/file_writer.py
```diff
@@ -38,6 +40,9 @@ def __init__(self, filename=None, mode='a', delimiter='\n', flush=True,
                      filename. By overridding time_format, other split
                      intervals, such as hourly or monthly, may be imposed.
 
+        split_interval Splits files based on a defined hour or minute time
```
davidpablocohn, line 43: Expand text a little to make it clear that suffix must be either 'H' or 'M'

logger/writers/file_writer.py
```diff
@@ -60,9 +65,55 @@ def __init__(self, filename=None, mode='a', delimiter='\n', flush=True,
...
         self.time_zone = time_zone
 
-        if split_by_time and not filename:
+        if (split_by_time or split_interval) is not None and not filename:
             raise ValueError('FileWriter: filename must be specified if '
-                             'split_by_time is True.')
+                             'split_by_time is specified or split_interval '
+                             'is True.')
+
+        if split_interval is None:
```
davidpablocohn, line 73: Get rid of conditional, and have
```
self.split_interval = None
self.next_file_split = 0
```
We want self.split_interval to be defined, regardless.

Then see below for checking whether it's time to roll over to a new file.

logger/writers/file_writer.py
```diff
@@ -60,9 +65,55 @@ def __init__(self, filename=None, mode='a', delimiter='\n', flush=True,
...
+            self.split_interval = None    
+
+        else:
```
davidpablocohn, line 76: Break the parsing of the interval off into its own little method, for clarity.  Maybe _split_interval_in_seconds()?

logger/writers/file_writer.py
```diff
@@ -120,7 +190,7 @@ def write(self, record):
 
         # If we're splitting by some time interval, see if it's time to
         # roll over to a new file.
-        if self.split_by_time:
+        if self.split_by_time or self.split_interval is not None:
```
davidpablocohn, line 193: More efficient - if splitting by interval, we only call _get_new_file_suffix() when the right amount of time has passed.
```
    if self.split_by_time or (self.split_interval and time.now() > self.next_file_split):
      if self.split_interval:  # if we're splitting by interval, set time for our next split
        self.next_file_split = self.next_file_split + self.split_interval 
    new_file_suffix = ...  # rest of clause
        
```

davidpablocohn (changes requested):
Can be more efficient with the interval checking by maintaining a "next_file_split" time variable - see inline.

logger/writers/file_writer.py
webbpinner: This is what I couldn't think of but figured was possible.  Makes much more sense.

davidpablocohn:
Yeah, that's why I get the big bucks... :)

BTW: this PR makes me finally understand the request from SailDrone in the
survey about "different logging intervals." Seems like a fine idea - thanks
for implementing it!

webbpinner:
Made changes based on recommendations.  These are untested but I wanted to make sure I was moving in the right direction.

davidpablocohn (approved):
Yup - this looks good.

webbpinner:
ok... I'll actually test the code this next round and will report back.

davidpablocohn:
Sounds good - thanks!

davidpablocohn (approved):

Issue #314, opened 2022-07-15
Feature Request: Option to add specify a header message when using FileWriter and LogFileWriter


webbpinner:
Would be useful to be able to specify a header message either as string or filepath (file would containing the header message) when using the FileWriter or LogFileWriter.  The defined header message would be added to the output file when the file is created.

PR #315, opened 2022-07-17
added ability to add headers to text_files, log_files & text-based files

webbpinner:
Code changes and update unit tests need to implement #314

### DIFF
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -17,7 +17,7 @@ class FileWriter(Writer):
     """Write to the specified file. If filename is empty, write to stdout."""
 
     def __init__(self, filename=None, mode='a', delimiter='\n', flush=True,
-                 split_by_time=False, split_interval=None,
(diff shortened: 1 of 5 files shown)


davidpablocohn (commented):
Thanks for this! Sanity checking: if I'm reading the logic correctly, the code looks at the `header` param string and checks whether it exists as a filename. If so, it uses the contents of the file, and if not, it it uses the header string itself as the header? It would be more parameters, but I'm torn on whether it would be less error-prone to have two separate (but mutually exclusive) parameters of `header` and `header_file`. Thoughts?

webbpinner:
Correct, ...and I was also torn between the current implementation vs having 2 header args.  In the end I would just like the ability to add files headers from a string or filepath.  I'm less concerned whether it's one or two args.  What makes the most sense to you and is most inline which the rest of the project?

davidpablocohn:
Looks good - thank you!

Issue #338, opened 2023-07-21
Add optional heading offset to truewind_transform


webbpinner:
Wind sensors are installed backwards with shocking frequency.  Request to add a heading offset to the truewind transform to account for this

davidpablocohn:
Already handled by "zero_reference_line" parameter.

PR #339, opened 2023-07-21
Issue 338

webbpinner:
Checking in feature request for a rel wind dir offset to account for misaligned wind sensors issue: #338


davidpablocohn (approved):

webbpinner:
So... my bad for not closely looking at the existing transform. The 'zero_line_reference' is (I'm assuming) how SAMOS defined the installation offset.  This pull request can be reverted and deleted.

PR #343, opened 2023-08-22
Reader hex string

ellapiet:
add option for bytes to be interpreted as hexadecimal but with string type (i.e. we are effectively not decoding the bytes, just changing the representation/type)


davidpablocohn (approved):
Looks good - thanks!

PR #344, opened 2023-08-22
generate yaml config with jinja

ellapiet:


davidpablocohn (approved):
Since this is in contrib, I'm not giving it a particulary deep dive. But looks good - thank you!

PR #350, opened 2023-10-25
Make FileWriter support raw/binary output and other fixes

veggiemike:
Hey, Pablo!  Here's a batch of fixes I made while making FileWriter work with raw/binary output.

### DIFF
--- a/logger/readers/reader.py
+++ b/logger/readers/reader.py
@@ -34,6 +34,12 @@ class Reader:
     def __init__(self, output_format=formats.Unknown,
                  encoding='utf-8', encoding_errors='ignore'):
         self.output_format(output_format)
+        # make sure '' behaves the same as None, which is what all the
+        # docstrings say, and would be logical... but then certain things treat
+        # them differently (e.g., file.open(mode='ab', encoding='') throws
+        # ValueError: binary mode doesn't take an encoding argument)
+        if encoding == '':
+            encoding = None
         self.encoding = encoding
         self.encoding_errors = encoding_errors
 
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -284,9 +284,6 @@ class FileWriter(Writer):
             if not self.file:
                 self._set_file(self.filename)
 
-        if isinstance(record, dict):
-            record = str(record)
-
         # Write the record and flush if requested
         self.file.write(record)
         if self.delimiter is not None:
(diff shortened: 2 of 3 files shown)


logger/readers/reader.py
```diff
@@ -34,6 +34,12 @@ class Reader:
     def __init__(self, output_format=formats.Unknown,
                  encoding='utf-8', encoding_errors='ignore'):
         self.output_format(output_format)
+        # make sure '' behaves the same as None, which is what all the
+        # docstrings say, and would be logical... but then certain things treat
+        # them differently (e.g., file.open(mode='ab', encoding='') throws
+        # ValueError: binary mode doesn't take an encoding argument)
+        if encoding == '':
```
davidpablocohn, line 41: Oops - thanks!

logger/writers/file_writer.py
```diff
@@ -52,19 +54,30 @@ def __init__(self, filename=None, mode='a', delimiter='\n', flush=True,
                      e.g. '-%Y-%m' is used, files will be split by month;
                      if -%y-%m-%d:%H' is specified, splits will be
                      hourly. If '%y+%j' is specified, splits will be
-                     daily, but named via Julian date.
+                     daily, but named via Julian date.  Putting '-' or '.' on
```
davidpablocohn, line 57: This is a great idea. At risk of getting too verbose, it may be worth adding that if the time format both ends *and* starts with '-' (or '.'), it will be used as a prefix. Actually thinking about it, filenames that start with a dash are problematic in Linux; would you mind flipping the code so that if it both starts and ends with a dash (or '.') it becomes a suffix?

logger/writers/file_writer.py
```diff
@@ -249,17 +278,20 @@ def write(self, record):
             new_file_suffix = self._get_file_suffix()
             if new_file_suffix != self.file_suffix:
                 self.file_suffix = new_file_suffix
-                self._set_file(self.filename + new_file_suffix)
+                if new_file_suffix.endswith('-') or new_file_suffix.endswith('.'):
+                    # make it a prefix even though we've called it a suffix
```
davidpablocohn, line 282: See note above about behavior when 'suffix' both starts and ends with a '-'/'.'; I *think* the more benign behavior would be to have it be used as a suffix rather than prefix, but am willing to be convinced.

davidpablocohn (changes requested):

logger/writers/file_writer.py
veggiemike: No convincing needed, that would indeed be the less ugly result.  I'll make it work that way

davidpablocohn (approved):

PR #352, opened 2023-10-29
Revamp UDPReader

veggiemike:
Hey, Pablo!  Happy weekend!  Don't read this until Monday, unless you really want to.  :-)  This batch of commits makes changes similar to the ones in recent UDPWriter and TCPWriter pull requests.  Improvements have been made to name resolution, text and raw/binary input handling, multicast and broadcast handling, optimized buffer sizes, configurable socket options, removal of TCP-centric conditional code, bet test coverage, etc.  This will be followed up with a "Replace NetworkReader with TCPReader" pull request once I've finished testing that code.

### DIFF
--- a/logger/listener/listen.py
+++ b/logger/listener/listen.py
@@ -690,7 +690,6 @@ if __name__ == '__main__':
                         addr.insert(0, '')
                     source = addr[0]
                     port = int(addr[1])
-                    print("encoding:", encoding)
                     readers.append(UDPReader(source, port, encoding=encoding))
 
             if new_args.redis:
--- a/logger/listener/listen.py
+++ b/logger/listener/listen.py
@@ -550,6 +550,7 @@ if __name__ == '__main__':
 
     LOG_LEVELS = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
     log_level = LOG_LEVELS[min(parsed_args.verbosity, max(LOG_LEVELS))]
+    logging.getLogger().setLevel(log_level)
 
     console_handler = logging.StreamHandler()
     console_handler.setFormatter(STDERR_FORMATTER)
--- a/logger/listener/listen.py
+++ b/logger/listener/listen.py
@@ -679,7 +679,6 @@ if __name__ == '__main__':
                     readers.append(NetworkReader(network=addr, encoding=encoding))
 
             if new_args.udp:
-                eol = all_args.network_eol
                 encoding = all_args.encoding
                 for addr_str in new_args.udp.split(','):
                     addr = addr_str.split(':')
@@ -690,7 +689,8 @@ if __name__ == '__main__':
                         addr.insert(0, '')
                     source = addr[0]
                     port = int(addr[1])
-                    readers.append(UDPReader(port=port, source=source, eol=eol, encoding=encoding))
+                    print("encoding:", encoding)
+                    readers.append(UDPReader(source, port, encoding=encoding))
 
             if new_args.redis:
                 for channel in new_args.redis.split(','):
@@ -827,8 +827,7 @@ if __name__ == '__main__':
                         addr.insert(0, '')
                     dest = addr[0]
                     port = int(addr[1])
-                    writers.append(TCPWriter(port=port, destination=dest,
-                                             eol=eol, encoding=encoding))
+                    writers.append(TCPWriter(dest, port, eol=eol, encoding=encoding))
 
             if new_args.write_udp:
                 eol = all_args.network_eol
@@ -842,8 +841,7 @@ if __name__ == '__main__':
                         addr.insert(0, '')
                     dest = addr[0]
                     port = int(addr[1])
-                    writers.append(UDPWriter(port=port, destination=dest,
-                                             eol=eol, encoding=encoding))
+                    writers.append(UDPWriter(dest, port, eol=eol, encoding=encoding))
 
             # SerialWriter is a little more complicated than other readers
             # because it can take so many parameters. Use the kwargs trick to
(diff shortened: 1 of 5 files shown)


logger/readers/udp_reader.py
```diff
@@ -8,56 +8,44 @@
...
 class UDPReader(Reader):
-    """
-    Read UDP broadcast and multicast records from a socket.
-    """
+    """Read UDP packets from network."""
     ############################
-
-    def __init__(self, port, source='', eol=None,
-                 read_buffer_size=READ_BUFFER_SIZE,
+    def __init__(self, interface, port, mc_group=None,
```
davidpablocohn, line 28: Need to also change listen.py:

(venv) rvdas@openrvdas-test:/opt/openrvdas$ logger/listener/listen.py --udp 6224
Traceback (most recent call last):
  File "/opt/openrvdas/logger/listener/listen.py", line 693, in <module>
    readers.append(UDPReader(port=port, source=source, eol=eol, encoding=encoding))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: UDPReader.__init__() got an unexpected keyword argument 'source'
(venv) rvdas@openrvdas-test:/opt/openrvdas$

logger/readers/udp_reader.py
```diff
@@ -70,70 +58,107 @@ def __init__(self, port, source='', eol=None,
...
-            # If no eol string, read, append, and try again.
-            record = self.socket.recv(self.read_buffer_size)
-            logging.debug('UDPReader.read() received %d bytes', len(record))
-            if record:
-                self.record_buffer += self._decode_bytes(record)
+        # If socket isn't ready, set it up.  If something fails, return w/out reading.
+        if not self.socket:
+            self.socket = self._open_socket()
+        if not self.socket:
+            logging.error('Unable to read record')
```
davidpablocohn, line 155: Maybe "Unable to open UDP socket for read"? - A little more specific about the source of the failure.

logger/readers/udp_reader.py
```diff
@@ -70,70 +58,107 @@ def __init__(self, port, source='', eol=None,
...
+        self.reuseport = reuseport
 
-        # If source is empty, we're listening for broadcasts, otherwise
-        # listening for multicast on IP 'source'
-        self.socket.bind((source, port))
+        # socket gets initialized on-demand in read()
+        self.socket = None
+
+    ############################
+    def _open_socket(self):
```
davidpablocohn, line 94: Good!

logger/readers/udp_reader.py
veggiemike: Oops, you're right.  I missed that.  Will push another commit out shortly.

logger/readers/udp_reader.py
veggiemike: Will do.

logger/listener/listen.py
```diff
@@ -690,7 +690,8 @@ def __init__(self, config_file, config_name=None, log_level=None):
                         addr.insert(0, '')
                     source = addr[0]
                     port = int(addr[1])
-                    readers.append(UDPReader(port=port, source=source, eol=eol, encoding=encoding))
+                    print("encoding:", encoding)
```
davidpablocohn, line 693: Is this "print()" left over from debugging?

logger/readers/udp_reader.py
davidpablocohn, line 28: Dang it - I realize that I've steered you wrong on UDPReader and UDPWriter: in order to not break scripts that are already out there in the world, we need 'interface' to have a default value, so you can define a UDPReader giving it just a port:

```
    - class: UDPReader
      kwargs:
        port: 6224
```
That construct is running in a lot of configs on a lot of ships.

This means interface needs a default value. And because Python doesn't let you have an arg without a default value following and arg that has one, that means that either 1) port also needs a default argument(??) or 2) we need port to be the first argument.

Harumph. Sorry for steering you astray here. Will think about whether I see an easy way forward - thoughts on a default UDP port? Or whether to just put up with the non-standardness of having port first?

davidpablocohn, line 28: Same issue as I've just flagged with UDPReader. Sorry! Thinking what way forward would be best, and more than happy to get suggestions.

davidpablocohn (commented):
Argh - see below. I've steered us into a compatibility vs code cleanliness corner. Would welcome your thoughts.

logger/listener/listen.py
veggiemike: Indeed it is.  I'll remove it.  And actually, I'll write up an Issue for why I was desperately trying to figure out what was going on right there...  Nothing major, can't decide if it's really an Issue or just plain old "user error".

logger/readers/udp_reader.py
veggiemike: Ah, yup, I see your point there.

I don't like the idea of a *real* default port.  I think our best compromise is to add `port=None` to the parameter list, then throw an exception if port wasn't specified.  So in Python's eyes, it's an optional kwarg, but in reality we make it required.  Then we can maintain the behavior in the .yml you just described and have a more natural dest,port pairing consistent across all the networking Readers/Writers.

Or, we reverse port/interface in UDPReader and TCPReader (once finished) so it's more Python-correct.

I'm gonna go implement the port=None workaround and see if I hate it.  Thoughts?

davidpablocohn:
Perfect - wish I'd thought of that!

davidpablocohn (approved):

PR #364, opened 2024-03-07
new functionality for nmea transform

ellapiet:
add choice for which nmeas to use, add dpt and stn nmeas, add test

### DIFF
--- a/logger/transforms/nmea_transform.py
+++ b/logger/transforms/nmea_transform.py
@@ -48,13 +48,25 @@ class NMEATransform:
         **kwargs
                 Arugments needed for the nmea transforms, see transforms below for what will be included.
         """
+        # If nmea_list is not given as list, force it into one
+        if not isinstance(nmea_list, list):
+            nmea_list = [nmea_list]
+
         class_module_name = 'logger.transforms.nmea_transform'
         module = importlib.import_module(class_module_name)
 
+        # Get all classes within this file
+        classes = [cls_name for cls_name, cls_obj in inspect.getmembers(module) if
+                   inspect.isclass(cls_obj)]
+
         self.transforms = []
+
         for transform in nmea_list:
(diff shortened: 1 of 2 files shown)


logger/transforms/nmea_transform.py
```diff
@@ -40,8 +41,21 @@ class NMEATransform:
...
+                List of the nmea transforms that will be used.
+        **kwargs
+                Arugments needed for the nmea transforms, see transforms below for what will be included.
+        """
+        class_module_name = 'logger.transforms.nmea_transform'
+        module = importlib.import_module(class_module_name)
+
+        self.transforms = []
+        for transform in nmea_list:
+            class_const = getattr(module, transform)
```
davidpablocohn, line 56: Sanity checking here - can we fail more gently if user specifies a transform that isn't defined? I'm guessing it's too messy to procedurally emit a message that says what the allowable transforms are, but perhaps we could refer them to the code?

davidpablocohn (changes requested):
I've not gone into depth on the the new transforms themselves, but in general they look reasonable. Only change I'd like would be some guardrails if the user specifies a transform that doesn't exist, so they can pinpoint what's wrong with their config.

Thanks!

logger/transforms/nmea_transform.py
ellapiet: Done! Added check to compare provided transforms with all classes from the nmea_transforms file

logger/transforms/nmea_transform.py
```diff
@@ -23,6 +23,7 @@
 # flake8: noqa E501  - ignore long comment lines that describe formats
 
 import logging
+import importlib
```
ellapiet, line 27: `import inspect`

davidpablocohn (approved):

PR #370, opened 2024-04-06
restores original api for nmea_transform

webbpinner:
the updated API breaks existing logger config files that use the NMEATransform.  This code makes the new nmea_list positional argument optional (default value = []).  In the absence of the nmea_list argument the transform reverts to it's previous API.


davidpablocohn (approved):

davidpablocohn:
Webb - could you sanity check that I did the merge correctly? (Sitting in campervan in the rain in Tassie, so not best bandwidth at the moment)

PR #378, opened 2024-04-29
Django Rest Framework - Code review

yoLewshi:
Pablo asked me to take a look at this. It all looks good to me just some minor tidying up, only change I would make is to group the endpoints which i've commented on below. 

I wasn't able to tag @gabesargeant for a review but obviously keen to know what you think as you did all the work so far!

### DIFF
--- a/django_gui/api_views.py
+++ b/django_gui/api_views.py
@@ -55,8 +55,7 @@ def api_root(request, format=None):
     These are an authenticated set of urls.
     You need to to login to use them via the APi ui.
     For calling them via curl or requests. You need to pass header authenticiation.
-    Use token auth, it's simpler.
-    But basic auth is supported.
+    Use token auth, it's simpler, but basic auth is supported.
     https://www.django-rest-framework.org/api-guide/authentication/
 
     Such as:
--- a/django_gui/api_views.py
+++ b/django_gui/api_views.py
@@ -15,7 +15,8 @@ from rest_framework.schemas import ManualSchema, coreapi as coreapi_schema
 from rest_framework.views import APIView
 
 from .views import log_request
-
+from django.urls import include, path
+from drf_spectacular.views import SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView
 import json
 import logging
 import yaml
@@ -44,11 +45,15 @@ from logger.utils.read_config import parse, read_config  # noqa: E402
 
 # Interacting with the Django DB via its API class
 # As per the standard view.
+
 api = None
 
+class GetSerializer(serializers.Serializer):
+    #Empty serializer that stops a warning from drf-spectacular ui
+    #Used  by api root, and get cruise config.
+    pass
 
-@api_view(["GET"])
-def api_root(request, format=None):
+class ApiRoot(APIView):
     """
     RVDAS REST API Endpoints.
 
@@ -69,9 +74,11 @@ def api_root(request, format=None):
     /api/schema/docs to use the openai version of this interface.
 
     """
-    return Response(
-        {
-            # flake8: noqa E501
+
+    serializer_class = GetSerializer
+    
+    def get(self, request, format=None):
+        return Response({
             "login": reverse("rest_framework:login", request=request, format=format),
             "logout": reverse("rest_framework:logout", request=request, format=format),
             "obtain-auth-token": reverse("obtain-auth-token", request=request, format=format),
@@ -81,8 +88,9 @@ def api_root(request, format=None):
             "delete-configuration": reverse("delete-configuration", request=request, format=format),
             "edit-logger-config": reverse("edit-logger-config", request=request, format=format),
             "load-configuration-file": reverse("load-configuration-file", request=request, format=format),
-        }
-    )
+        })
+
+
 
 
 def _get_api():
@@ -124,7 +132,7 @@ class CustomAuthToken(ObtainAuthToken):
 class CruiseConfigurationAPIView(APIView):
     authentication_classes = [authentication.BasicAuthentication, TokenAuthentication]
     permission_classes = [IsAuthenticated]
-
+    serializer_class = GetSerializer
     def get(self, request):
         log_request(request, "get_configuration")
         template_vars = _get_cruise_config()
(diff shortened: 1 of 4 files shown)


django_gui/api_views.py
```diff
@@ -0,0 +1,578 @@
...
+@api_view(["GET"])
+def api_root(request, format=None):
+    """
+    RVDAS REST API Endpoints.
+
+    These are an authenticated set of urls.
+    You need to to login to use them via the APi ui.
+    For calling them via curl or requests. You need to pass header authenticiation.
+    Use token auth, it's simpler.
+    But basic auth is supported.
```
yoLewshi, line 59:
```suggestion
    Use token auth, it's simpler, but basic auth is supported.
```

django_gui/urls.py
```diff
@@ -13,33 +13,49 @@
...
+    path('api/obtain-auth-token/', api_views.CustomAuthToken.as_view(), name="obtain-auth-token"),
+    # These are the DRF API stubbs
+    # flake8: noqa E501
+    path('api/', api_views.api_root),
+    path('api/cruise-configuration/', api_views.CruiseConfigurationAPIView.as_view(), name='cruise-configuration'),
+    path('api/delete-configuration/', api_views.CruiseDeleteConfigurationAPIView.as_view(), name='delete-configuration'),
+    path('api/edit-logger-config/', api_views.EditLoggerConfigAPIView.as_view(), name='edit-logger-config'),
+    path('api/load-configuration-file/', api_views.LoadConfigurationFileAPIView.as_view(), name='load-configuration-file'),
+    path('api/reload-current-configuration/', api_views.CruiseReloadCurrentConfigurationAPIView.as_view(), name='reload-current-configuration'),
+    path('api/select-cruise-mode/', api_views.CruiseSelectModeAPIView.as_view(), name='select-cruise-mode')
```
yoLewshi, line 60: I would prefer to have the routes grouped by the model/area they're dealing with, something like this:

```suggestion
    path('api/cruise/get-config/', api_views.CruiseConfigurationAPIView.as_view(), name='cruise-configuration'),
    path('api/cruise/delete-config/', api_views.CruiseDeleteConfigurationAPIView.as_view(), name='delete-configuration'),
    path('api/cruise/reload-current-config/', api_views.CruiseReloadCurrentConfigurationAPIView.as_view(), name='reload-current-configuration'),
    path('api/cruise/select-mode/', api_views.CruiseSelectModeAPIView.as_view(), name='select-cruise-mode'),
    path('api/load-config-file/', api_views.LoadConfigurationFileAPIView.as_view(), name='load-configuration-file'),
    path('api/logger/edit-config/', api_views.EditLoggerConfigAPIView.as_view(), name='edit-logger-config')
```

django_gui/api_views.py
```diff
@@ -0,0 +1,577 @@
...
+            data["active_mode"] = api.get_active_mode()
+        except (ValueError, AttributeError):
+            logging.info("No configuration loaded")
+            data = {"modes": "no cruise loaded"}
+
+        return Response({"status": "ok", "data": data}, 200)
+
+
+class CruiseReloadCurrentConfiguartionSerializer(serializers.Serializer):
+    # The presence of the value implies the
```
yoLewshi, line 232: what would the rest of this comment be?

django_gui/urls.py
```diff
@@ -13,33 +13,49 @@
...
-    path('', views.index, name='index'),
+    #
+    #API Dango REST Framework Views
+    #
+    # DRF Spectacular urls.
+    path('api/schema/', SpectacularAPIView.as_view(), name='schema'),
+    path('api/schema/swagger-ui/', SpectacularSwaggerView.as_view(url_name='schema'), name='swagger-ui'),
+    path('api/schema/redoc/', SpectacularRedocView.as_view(url_name='schema'), name='redoc'),
+    # These are the login and logout restframework urls
+    path('api/', include('rest_framework.urls', namespace='rest_framework')),
```
yoLewshi, line 49: I'm torn between whether it is cleaner to have all the other methods defined elsewhere and imported like these are with `include()` or if it's better to define all the routes in this file. 

It's harder to integrate upstream changes if you're adding all your custom endpoint routes here, I already have custom endpoints that i'm adding to this by concatenating (` + niwa_views.get_paths()`) and I think we should define a pattern for this. Perhaps if we set these rest framework ones out it can become the blueprint for anyone wanting to add their own endpoints. I'm thinking it would look something like:

```
path('api/', include('api_views.cruise.urls', namespace='cruise')),
path('api/', include('api_views.logger.urls', namespace='logger'))
```

(not sure on the syntax needed)

django_gui/urls.py
gabesargeant: Hi, 
Nice change, I all for this approach it's a cleaner setup.

django_gui/api_views.py
gabesargeant: That comment should be something along the line of. 
The presence of the value implies true, as the implementation in the the main views.py at line: ~93 
switches on the presence of a form element. For consistency, the input object is expecting json {"reload": "true"} but the reload key is really the switch.

django_gui/urls.py
gabesargeant: I like this suggestion. I would just keep both cruise and logger urls together as a set of 'core' api urls even if they were split by /logger/cruise as you previously suggested. 

I think that just grouping the core api's as a set of urls that exist in a in the api_views would be enough separation for what exists now in the codebase. 

If you have a look at path('admin/', admin.site.urls), the urls method there is invoked at runtime by the django frame to provide the admin urls. That's a pretty neat looking pattern to follow. 

I think having users integrate their changes via a django 'app' is probably the best integration pattern for the django components.

yoLewshi: Sounds good to me, are you able to make the change on this branch? I'm not familiar enough with Django apps or the rest framework to know whether I've done it correctly.

gabesargeant: Hi, 
Have a glance at 
https://github.com/OceanDataTools/openrvdas/compare/issue_372...gabesargeant:openrvdas:issue_372
those changes are pretty simple. 
WRT to your niwa_views.get_paths()
If you defined your paths in a urlpatterns list, then you can do a similar import to like what I have with the API. 
I found this a pretty good read on the topic. 
https://docs.djangoproject.com/en/5.0/topics/http/urls/
Cheers.

davidpablocohn:
Thanks, Gabe and Lewis!

I've been pulling these changes incrementally into the issue_372 branch,
but am realizing that that's probably been a bad idea, as I've lost track
of what functionality I've merged in or not, and how close to convergence
you two are on whether the code is ready for prime time. Low urgency, but
could one of you give me a sanity check on where we are?

Thanks again,
 -p

davidpablocohn (approved):
Merging into dev to pressure test

PR #397, opened 2024-09-19
regex record parser with example and test

ellapiet:
Here is our regex record parser. Set up as a super class of the standard record parser and overwriting a few functions. With the way that regex capturing groups work we also needed a slightly different way to define data types, so we've included that in the fields meta data section of each device definition.


davidpablocohn (approved):
Looks good to me, thanks! Apologies for the delay in getting to this.

PR #401, opened 2024-10-10
Refactor of Transform class to handle cross cutting concerns

mgeard:
…oss-cutting concerns

* Transform class renamed to BaseTransform to better reflect its purpose
* BaseTransform's transform method now handles None check and recursion cross-cutting concerns
* Most concrete transform classes now override the _transform_single_record method, with recursion and None checks handled by the base transform

Some concrete transforms still override the transform method rather than the new _transform_single_record
* TimestampTransform, due to the ts param
* InterpolationTransform, due to its unique _add_record logic
* NMEATransform, due to its iterative logic

### DIFF
--- a/logger/transforms/base_transform.py
+++ b/logger/transforms/base_transform.py
@@ -7,7 +7,7 @@ from logger.utils import formats  # noqa: E402
 
 
 ################################################################################
-class Transform:
+class BaseTransform:
     """
     Base class Transform about which we know nothing else. By default the
     input and output formats are Unknown unless overridden.
@@ -44,7 +44,15 @@ class Transform:
 
     ############################
     def transform(self, record):
-        """Should return None if the result of transformation is empty record"""
-        raise NotImplementedError('Class %s (subclass of Transform) is missing '
-                                  'implementation of transform() method.'
+        if record is None:
+            return None
+        if isinstance(record, list):
+            return [self.transform(single_record) for single_record in record]
+
+        return self._transform_single_record(record)
+
+    ############################
+    def _transform_single_record(self, record):
+        raise NotImplementedError('Class %s (subclass of BaseTransform) is missing '
+                                  'implementation of _transform_single() method.'
                                   % self.__class__.__name__)
--- a/logger/transforms/count_transform.py
+++ b/logger/transforms/count_transform.py
@@ -7,12 +7,12 @@ from os.path import dirname, realpath
 sys.path.append(dirname(dirname(dirname(realpath(__file__)))))
 from logger.utils import formats  # noqa: E402
 from logger.utils.das_record import DASRecord  # noqa: E402
-from logger.transforms.transform import Transform  # noqa: E402
+from logger.transforms.base_transform import BaseTransform  # noqa: E402
 
 
 ################################################################################
 #
-class CountTransform(Transform):
+class CountTransform(BaseTransform):
     """Return number of times the passed fields have been seen as a dict
     (or DASRecord, depending on what was passed in) where the keys are
     'field_name:count' and the values are the number of times the passed
@@ -33,20 +33,8 @@ class CountTransform(Transform):
         self.counts = {}
 
     ############################
-    def transform(self, record):
+    def _transform_single_record(self, record):
         """Return counts of the previous times we've seen these field names."""
-        if not record:
-            return None
-
-        # If we've got a list, hope it's a list of records. Recurse,
-        # calling transform() on each of the list elements in order and
-        # return the resulting list.
-        if type(record) is list:
-            results = []
-            for single_record in record:
-                results.append(self.transform(single_record))
-            return results
-
         if type(record) is DASRecord:
             fields = record.fields
         elif type(record) is dict:
--- a/logger/transforms/delta_transform.py
+++ b/logger/transforms/delta_transform.py
@@ -1,5 +1,6 @@
 import logging
 
+from logger.transforms.base_transform import BaseTransform
 from logger.utils.das_record import DASRecord  # noqa: E402
 
 KNOWN_FIELD_TYPES = ['polar']
@@ -11,7 +12,7 @@ def polar_diff(last_value, value):
 
 
 ################################################################################
-class DeltaTransform:
+class DeltaTransform(BaseTransform):
     def __init__(self, rate=False, field_type=None):
         """Return a DASRecord (or dict, depending on input record type) with
         each field's delta in value from it's previous value. If a field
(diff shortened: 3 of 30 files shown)


davidpablocohn:
Hi Markus - thanks for this!

I'm scrambling at the moment to get presentations ready for RVTEC, so it's
going to be a few days before I'm able to give this my attention.

I love the idea of consolidating the common multi-record processing into a
single parent class. But I'm going to need to sit with the idea of
Transforms no longer having their own transform() method. The code itself
looks like a solid implementation - it's mostly the change in nomenclature
that I need to think about. I'll try to get back to you with more thoughts
mid-week.

Thanks again,
 -p

(Also: I see you're in Brisbane - are you with the AAD?)

mgeard:
Hey Pablo,
Cheers for the response.

Transform concrete classes may still have have their own transform() method if one chooses to override it - See the TimestampTransform as an example.

One alternative approach that I was contemplating was an aspect oriented one. This approach would allow concrete classes to override transform() directly.

Here's what such an implementation may look like:


```
class ConcreteTransform(BaseTransform)

@none_check_aspect
@recursion_aspect
def transform()
    # logic for transform goes here
```



with the aspects defined as such:

```
def none_check_aspect(func):
    @wraps(func)
    def wrapper(self, record):
        if record is None:
            return None
        return func(self, record)
    return wrapper

def recursion_aspect(func):
    @wraps(func)
    def wrapper(self, record):
        if isinstance(record, list):
            return [func(self, single_record) for single_record in record]
        return func(self, record)
    return wrapper
```

While the aspect oriented approach is a much more powerful solution it may come with the trade-off that it's more complicated for less experienced engineers to wrap their heads around; I have been told that this is a consideration that I may have to make when working with RVIDAS.

I'd love to hear your thoughts.

- Cheers


I'm with CSIRO in Hobart - Just realised I haven't updated my location in a while!

davidpablocohn:
Hi Markus - apologies for the continued radio silence on this! I'm still on
the road and, while I keep revisiting the code, haven't been able to come
up with a solid conclusion on whether this feels like the right way to take
it.

Also, thank you for the time and careful thought you've put into this.
You're spot on (about my perceptions, at least) that we need to keep
structures simple enough to not scare away folks who have fairly
rudimentary only-learned-what-I-need-to-learn Python skills. For that
reason I'd shy away from using decorators, as you've noted.

For much that same reason on a smaller scale, I'm feeling hesitant to lift
the transform() method out of the concrete classes. It makes absolute sense
to have the parent class contain code for dicing up lists of records, but
I'm leaning more toward having that be explicitly called from the concrete
class

def transform(self, record):
    if record is None:
        return None
    if isinstance(record, list):
        return self.list_transform(record)  # in parent class

    # Else do normal transform stuff
    ...

Then have the parent class have the list comprehension and cleanup (that
is, get rid of all Nones and, if the list is ends up empty, just reduce it
to a single None. Something like:

def list_transform(self, record):
    # Should test that we've actually got a list here..

    transformed = [self.transform(single_record) for single_record in
record]
    reduced = [r in transformed if r is not None]
    if not len(reduced):
        return None
    return reduced

davidpablocohn:
Hi Markus (and Webb),

Spent some more time on this, and here's where I'm leaning at the moment.
It would both digest lists of records and give the option of somewhat
uniform runtime type checking with minimal overhead.

*Abstract parent class* with is_simple_record() and digest_record() methods:

#############################
# Abstract parent class for transforms
class Transform:
    def __init__(self, input_types=None, quiet=False):
        self.input_types = input_types
        self.quiet = quiet
        self.class_name = self.__class__.__name__

        print('init Transform()')

    # Is this record in a format that the transform can handle?
    def is_simple_record(self, record):
        try:
            return record is None or isinstance(record, self.input_types)
        except AttributeError:
            class_name = self.__class__.__name__  # weren't
initialized, so need to do here
            logging.warning(f'{class_name} code error:
Transform(input_types) not initialized!')
            return True  # if not initialized, not clear we can digest

    # Try to digest record down into a format that the transform can handle.
    # Typically that will mean that we've been handed a list of records that
    # we need to break into individual records.
    def digest_record(self, record):
        try:
            logging.debug(f'trying to digest record "{record}"')
            logging.debug(f'needs to be of type {self.input_types}')
        except AttributeError:
            class_name = self.__class__.__name__  # weren't
initialized, so need to do here
            logging.warning(f'{class_name} code error:
Transform(input_types) not initialized!')
            return record  # if not initialized, not clear we can digest

        # If it's a type the transform can handle directly (though, if so,
        # why were we called?!?
        if self.is_simple_record(record):
            logging.warning(f'{self.class_name}: digest_record()
called with simple record.')
            logging.warning(f'Received {type(record)}: "{record}"')
            return self.transform(record)

        # We know how to deal with it if it's a list: Apply to components,
        # stripping out any None's
        if isinstance(record, list):
            return [self.transform(r) for r in record if r is not None]

        # If we don't know how to deal with it
        if not self.quiet:
            logging.warning(f'Unable to convert record to format
"{self.class_name}" can process')
            logging.warning(f'Must be instance or list of {self.input_types}')
            logging.warning(f'Received {type(record)}: "{record}"')
        return None

*Concrete classes* call super().__init__() to enable methods of
abstract class, and use them to digest records when necessary:

#############################
# Concrete transform class

class ChildTransform(Transform):
    def __init__(self):
        super().__init__(input_types=(int, str, float))  # types of
records we can handle
        print('init child')

    def transform(self, record):
        # If record is not of a type we know how to process, see if we can
        # digest it into something we *can* process, then call transform()
        # recursively.
        if not self.is_simple_record(record):  # inherited from Transform()
            return self.digest_record(record)  # inherited from Transform()

        # Do the actual transformation
        result = (Whatever our transform is...)
        return result

Would love to hear what y'all think before I start doing any surgery
to put this in place of the currently disused type checking.

Thoughts?

mgeard:
Hey Pablo,
It seems like a pretty well structured compromise to me.

While I do have a strong love for aspect oriented designs, I agree that it may not be the most appropriate here. The solution you have provided is KISS. 

The methods is_simple_record _could_ be clearer but I don't know what to suggest; It's just a hunch I have.

It looks good.

Thanks,
M

davidpablocohn:
Agreed on is_simple_record() - I'm not even sure it's got the right name
yet. I'll keep mulling it for a bit before I start breaking ground on new
code.

Thanks, and I'll keep you posted!

-p

(BTW: my wife has a business trip to NZ in February, and I'm planning to
tag along and add some time in Tassie when she's done, including a couple
of days in Hobart. I've been keeping Ella in the loop on that planning, and
if it works out, I'll hopefully get to meet you in person and thank you for
your contributions!)

davidpablocohn:
Note that I've started working on this in #410. After a false start, I realized that instead of having the class explicitly hand in a list of types, I should look up and make use of the type hints mechanism that is already in Python. Behind the scenes coding is a bit messier, but it will result in a more Pythonic result.

PR #437, opened 2025-04-10
Tried to add config_templates to read_config but failed

webbpinner:
I submit and admit defeat...  Best I could do was build the templates and new config files


davidpablocohn (approved):

davidpablocohn:
Okay - I think I got it. A little manual smooshing, and a little help from
Claude.ai, I generalized expand_logger_templates and
expand_config_templates into a single method.

Can you have a look at branch_422 when you get the chance, and see if it
looks right to you?

Issue #422, opened 2025-02-23
Allow (parameterized) reusable logger definitions in cruise definition file


davidpablocohn:
Much the way devices and device types are reusable, the overhead of creating and maintaining a cruise definition file could be simplified if it was possible to define a few parameterized logger types and reuse them. 

It would operate along the lines of jinja, running a preprocessor over the initial file (perhaps automatically when the templatized file is loaded?) to expand it. We discussed some syntax conventions that seem like they would make this scale better than raw Jinja.

![Image](https://github.com/user-attachments/assets/e2ba1501-0fda-45a9-8224-a0c06cc72be9)

webbpinner:
My thoughts...
Combining config definitions into the loggers section makes sense.

Field name patterns for devices makes sense... I personally would prefer the default pattern to be %device%_%field%

Still on the fence with templates... I find the yaml native variable expansion to be a real improvement over what I've done in the past.  That said... I guess one way to implement would be:

Here's what goes in logger_config.yaml:
```
---
configs:
  cwnc-net:
    template: serial_to_net
      kwargs:
        prefix: %prefix%
        baudrate: %baudrate%
        serial_port: %serial_port%
```

Here's the template... written in it's own file(s) that are included with the logger_config via 'includes' OR written inline within the logger_config.yaml within a `templates` section:
```
---
templates:
  serial_to_net:
    readers:
    - class: SerialReader
      kwargs:
        baudrate: %baudrate%
        port: %serial_port%
    transforms:
    - class: TimestampTransform
    - class: PrefixTransform
      kwargs:
        prefix: %prefix%
    writers:
    - class: UDPWriter
      kwargs:
        port: 6224
        destination: 255.255.255.255
```

So the complete logger_config.yaml would look something like:
```
---
########################################
loggers:
  PCOD:
    configs:
    - PCOD-off: {}
    - PCOD-net:
       template: serial_to_net
          kwargs:
             prefix: PCOD
             baudrate: 4800
             serial_port: /tmp/tty_PCOD
  cwnc:
    configs:
    - cwnc-off: {}
    - cwnc-net:
       template: serial_to_net
          kwargs:
             prefix: cwnc
             baudrate: 19200
             serial_port: /tmp/tty_cwnc

########################################
templates:
  serial_to_net:
    readers:
    - class: SerialReader
      kwargs:
        baudrate: %baudrate%
        port: %serial_port%
    transforms:
    - class: TimestampTransform
    - class: PrefixTransform
[...]

PR #439, opened 2025-04-12
Issue 422

davidpablocohn:
Add in config_templates, as per #422  discussion.


webbpinner:
I took a look at the results and everything looked correct to me. I'm going to try to use it to build out a full ship deployment tomorrow.

davidpablocohn:
Excellent. I'm about to merge issue_422 into dev, so you can use that
branch.

PR #447, opened 2025-05-04
Resolves #446 - Dynamically build cds-ws URL to account for accessing OpenRVDAS via port forwarding

### LINKED ISSUE #337: Allow ws access when remotely accessing openrvdas webUI via port forwarding
Several vessels provide remote access to the OpenRVDAS via ssh-based port forwarding.  This works for the static portions of the OpenRVDAS webUI but because the port used for the ws connection is explicitly defined to a port number within the  settings.py file the ws socket traffic is not accessible when accessing via ssh port-forwarding.

Running on the assumption that the WS port is the same as the port used by nginx during install would it be possible to define the ws port as the same port specified in the URL.

TL;DR
Can the default vault of `WEBSOCKET_SERVER` be changed from:
```
var WEBSOCKET_SERVER = "ws://:80/cds-ws";
```
To:
```
var WEBSOCKET_SERVER = (document.location.protocol === 'https:' ? 'wss' : 'ws' ) + '//:' + document.location.port + "/cds-ws";
```

### LINKED ISSUE #446: Dynamcally build cds-ws URL to account for accessing OpenRVDAS via port forwarding
Accessing OpenRVDAS via port forwarding doesn't work because the CDS websocket URL is statically defined.

This could be made dynamic based on the django gui URL with the following patch:
```
--- a/django_gui/static/django_gui/websocket.js
+++ b/django_gui/static/django_gui/websocket.js
@@ -28,6 +28,8 @@ var websocket_server = WEBSOCKET_SERVER;
 if (websocket_server.indexOf('//:') > 0) {
   websocket_server = websocket_server.replace('//:', '//' + location.hostname + ':');
 }
+const url = new URL(websocket_server);
+websocket_server = url.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '') + url.pathname;
```

davidpablocohn:
This also resolves #337, right?

### DIFF
--- a/django_gui/static/django_gui/websocket.js
+++ b/django_gui/static/django_gui/websocket.js
@@ -29,7 +29,7 @@ if (websocket_server.indexOf('//:') > 0) {
   websocket_server = websocket_server.replace('//:', '//' + location.hostname + ':');
 }
 const url = new URL(websocket_server);
-const port_spec = window.location.port ? ':' + window.location.port : ''
+const port_spec = window.location.port ? ':' + window.location.port : '';
 websocket_server = url.protocol + '//' + window.location.hostname + port_spec + url.pathname;
 
 //////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


webbpinner:
Yes... this is a repeat of #337.  Thank you for the patch.

webbpinner (approved):
This looks correct to me.

Issue #454, opened 2025-07-23
Extension of serial_reader that behaves in accordance with WHOI's preferred format


webbpinner:
WHOI uses a bespoke data prefix consisting of a prefix string, timestamp in `%Y-%m-%d %H:%M:%S` format and optional sensor ID.  WHOI also has an established practice of logging empty data records.

The default serial reader does not support this format natively, thus a WHOI-specific extension to the default serial reader is required.

davidpablocohn:
Note: things are a bit crazy this morning, but I can try to have a proof-of-concept by this evening, to compare.

webbpinner:
This can probably be closed

davidpablocohn:
Solved via flag in #456

PR #456, opened 2025-07-25
#454 - added keep_empty_records flag

davidpablocohn:
Low urgency, but here's my interpretation of the change needed to fit WHOI requirements.


webbpinner:
also... is it worth adding the same logic to UDPReader while this is being worked on?  I guess technically that would be it's own feature request so maybe not.

davidpablocohn:
Agreed on both fronts - "allow_empty" and also adding to UDPReader. Will
get on that, thanks!

webbpinner (approved):

PR #461, opened 2025-08-19
Issue 379

webbpinner:
Initial import of SealogReader.  Please review and comment but hold off on accepting until I add the SealogWriter


webbpinner:
Ok... please take a look an ensure i didn't do anything stupid.  This code is being used (and working) on R/V Hydra

davidpablocohn (commented):
This looks good, thanks!

Do you see any situation where one might do any inside-OpenRVDAS manipulation of Sealog record after it was created? I don't feel strongly, but if not, my inclination would be to fold the ToSealogTransform into SealogWriter. 

At the least, I'd have a SealogWriter be able to (optionally) take a DASRecord and implicitly make the conversion internally to SealogRecord. Not having that will mean that you may end up having to set up extraneous ComposedWriters to encapuslate that transform+write.

webbpinner:
That was exactly why I asked.  The ComposedWriter issue   Started to refactor, will test tomorrow.

webbpinner:
Ok... I think this is in a decent place.  Refactored so that the SealogWriter works with or without the to_sealog_transform.  Expanded the sealog_config definition so that there's less need to manipulate the event object before submission to Sealog-Server.

webbpinner:
Not to be outdone by geofence_transform... I added the sealog_control_transform with enables users to send set_active_mode commands to based on sealog events.

PR #462, opened 2025-08-29
Issue 458

webbpinner:
First stab at this. Please review and provide comments...  Introduced several breaking changes for existing user configs but I think it's a step in the right direction for unifying the two classes.

### DIFF
--- a/logger/writers/file_writer.py
+++ b/logger/writers/file_writer.py
@@ -3,7 +3,9 @@
 import logging
 import os.path
 import sys
+import re
 import math
+import warnings
 
 from datetime import datetime, timedelta, timezone
 from typing import Union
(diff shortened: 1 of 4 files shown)


davidpablocohn:
Just starting to dig into this. Is the Sealog stuff inextricably linked to the file_writer changes? If not, I'd really like to have them as separate pulls. I've got great trepidation about any potentially-breaking changes, especially with FileWriter. I'd much rather introduce config versioning, with the config including a flag that says what format to expect.

davidpablocohn:
Oh, I've not gotten anywhere near seeing whether it works - just starting to get my head around the whole thing (coming up from a few weeks of non-code)

davidpablocohn:
Understood. I think my preferred path would be to do it in several steps: first enable the new calling convention, retaining the old, without a deprecation warning. At a later time, enable the warning, and then at some point long after that, disable the old.

I think what would help me here is having sample invocations in the docstring, showing how the different invocations would work. Yeah, I know we don't have them on any of the other writers, but I'd like us to.

webbpinner:
I've added the code requested changes... will add the docstring updates soon

webbpinner:
Oh the rabbit holes you lead me down... the updated code didn't account for j-day date_formats in it's validation.  Took care of that and added some additional tests.

PR #467, opened 2025-10-04
Issue 457

davidpablocohn:
Add GoogleSheetsWriter


webbpinner (approved):
Interesting addition.  I'm curious to know who requested this writer and their intended use-case.

davidpablocohn:
Self-requested. :-)

Intended use is for low-frequency home monitoring. Wondering if it should
be in some other location, like maybe somewhere in contrib.

PR #483, opened 2025-12-10
Replaces subprocess git tag retrieval with GitPython

webbpinner:
Uses GitPython library to retrieve the current git tag. This change addresses potential issues related to relying on shell commands, and improves error handling. Additionally, it adds GitPython as a dependency.


davidpablocohn:
I'm torn about this. I agree that avoiding a call to subprocess is good,
but it adds a dependency on another module that will need to get installed
(either manually or by re-running install_openrvdas.sh).

webbpinner:
True... but I typically re-run the install script any time I do an update like with 2.2's UI refresh.  The code includes an error message in the exception explaining that the dependency is missing and how to install it.  This code also returns additional repo metadata as an object so it's easier to integrate with the django code.

ellapiet:
I also deliberated this and didn't want to add the dependency. Do we have any other uses for gitpython? Otherwise I think staying with subprocess might be the go.

webbpinner:
Ported code back to using subprocess but retained the output object structure.  Since the output logic structure contains more than just the tag I updated the function name from `get_current_git_tag()` to `get_git_info()`. I'm realizing now it might more since to rename the file from `git_tag.py` to `git_info.py`.  Also added an `if __name__ == "__main__":` definition so the file can be called directly. 

Output structure:
```
{'tag': 'a2.0.5-34-g2144eb7', 'remote_url': 'https://github.com/OceanDataTools/openrvdas.git', 'branch': 'issue_479', 'commit': '2144eb78dbf9f2172ef4cf7b4f73e07ba25d07be'}
```

ellapiet (approved):
LGTM

PR #492, opened 2025-12-27
Issue 317

webbpinner:
After studying the GrafanaLiveWriter I refactored the HTTPReader to use native urllib vs the requests library to remove it as a dependency.


davidpablocohn (approved):
Looks good - thanks!

PR #528, opened 2026-03-07
Issue 523

davidpablocohn:
Remove local (out-of-sync) .md documentation files and replace with pointers to current docs in openrvdas-docs repo, along with instructions for installing and displaying docs locally. Resolves #523


webbpinner (approved):
LGTM... makes me realized I probably need to evaluate the sealog/openvdm docs.

PR #541, opened 2026-03-27
fix multicast on reader and writer

ellapiet:
Both: Added checking for the multicast IP range

UDP Reader: Added membership for the mc_group, then bound to '' as we have already joined. Also don't need to specify any other random destination.

UDP Writer: Generally speaking, send() is used for TCP SOCK_STREAM connected sockets, and sendto() is used for UDP SOCK_DGRAM unconnected datagram sockets. (https://www.gta.ufrj.br/ensino/eel878/sockets/sendman.html) So changed both cases to use sendto, now we don't need to add membership or connect to the socket

### DIFF
--- a/logger/readers/udp_reader.py
+++ b/logger/readers/udp_reader.py
@@ -3,6 +3,7 @@
 import logging
 import socket
 import sys
+import ipaddress
 
 from os.path import dirname, realpath
 sys.path.append(dirname(dirname(dirname(realpath(__file__)))))
@@ -55,6 +56,7 @@ class UDPReader(Reader):
         port         Port to listen to for packets.  REQUIRED
 
         mc_group     If specified, IP address of multicast group id to subscribe to.
+                    Must be in range 224.0.0.0 to 239.255.255.255
 
         reuseaddr    Specifies wether we set SO_REUSEADDR on the created socket.  If
                      you don't know you need this, don't enable it.
@@ -96,14 +98,13 @@ class UDPReader(Reader):
         if mc_group:
             # resolve once in constructor
             mc_group = socket.gethostbyname(mc_group)
+
+            if not ipaddress.ip_address(mc_group).is_multicast:
+                logging.error("mc_interface %s IPV4 Multicast addresses use reserved class D address range: 224.0.0.0 through 239.255.255.255", mc_group)
+                raise TypeError('must specify valid mc_interface')
+
             if not interface:
-                # multicast needs to specify interface to use, so let's pick a
-                # sane default
-                #
-                # NOTE: This means hostname cannot be an alias to localhost, or
-                #       you won't be able to send IGMP packets correctly.
-                #
-                self.interface = socket.gethostbyname(socket.gethostname())
+                self.interface = ''
 
         self.mc_group = mc_group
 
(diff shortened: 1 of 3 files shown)


webbpinner:
@davidpablocohn do you think it's worth having @veggiemike from OET take a look at this?  I believe he did a lot of the initial MC development

veggiemike:
> @davidpablocohn do you think it's worth having @veggiemike from OET take a look at this? I believe he did a lot of the initial MC development

Thanks, @webbpinner.  I'd love to take a look at this, but it'll have to wait until Monday.

davidpablocohn:
Thanks, all! I've had a look and it looks like a great fix, but I'm going to defer to @veggiemike to have a look on Monday as a sanity check

veggiemike:
I would not merge this as-is.  There's way too much going on in here other than fixing what's broken.  Speaking of that, I see no explanation of what the problem you're trying to fix actually was...  I can only assume it's that test_udp_writer was failing to join the IGMP group (because that's what I see when I run it).  I'm pretty sure that
failure on my laptop is due to my home WiFi setup (or maybe WiFi in general?), and we absolutely should fix this so all the tests pass.  The simple fix for that, buried in your first patch, is to switch to `INADDR_ANY` for the source IP in the tester instead of trying to autodetect a valid IP address to use for IGMP membership requests.  When I wrote the test, I was trying to make sure it tested as much as possible, but I'd say I was probably trying too hard, given that it's all going over local loopback anyways.

With that out of the way, here's some friendly feedback on submitting pull requests in general:

For starters, I'd break these commits up into commits that each do exactly one thing.  So for your pull request, I'd expect something like this:

- add ipaddress checks for multicast addresses (one commit for both
  reader and writer)
- use mreq instead of python concat for mc joining (in writer)
- use sendto instead of send for UDP writer
- use INADDR_ANY in udp testers (one commit for both read/write testers)

That makes code review (and maintenance) a ton easier, as we can have easy conversations like "Hey, I like the last commit, but I really don't like the sendto one".  We have the option to cherry-pick single parts of your request then, too.  Or revert them later, if needed.

It also makes it easier to make sure that each thing you're trying to do is COMPLETE, and if you tell me in words in the commit message what you're trying to accomplish I can easily verify that your code does exactly that.

You absolutely have to check for dead-code or lies in output or comments resulting from your changes.  For example, you've replaced the `source_ip` variable with `dest` (which by the way is completely backwards), but didn't remove `source_ip`, and it's still used in logging output which is now a lie.

Don't rename variables w/out a good reason: it makes the diffs harder to read for no gain.  The ones I see you changing aren't even any more correct than what's there.  For example, changing `host` to `mc_interface`.  Fair it's not a host but it's also not a multicast interface, it's an IP address.  I used host/dest to stay common with other variable naming in other testcode, to make comparisons easier, despite it not technically being a host.

Now onto the technical review comments:

I like the ipaddress check additions you made.

I'd strongly advise against removal of any of the extra error-checking in the actual reader/writer code (e.g., the localhost IGMP warning).  In fact, I'd I'd honestly rather not touch a single thing in the reader/writer w/out a really good reason, as I tested this thoroughly before submitting, rely on it working as intended for operations on the Nautilus, and only the testcode appears less than correct to me.  I don't have the time to fully re-test the multicast code today, and I'd absolutely want to prior to trusting any of the changes.  That being said, I'll change my tune in a heartbeat if there's an actual bug you're running into but didn't explicitly mention.

Regarding send vs sendto, you're half right.  In general sendto() is used for UDP, but not always, and send() is perfectly legal.  I chose to use send() instead of sendto() for the UDP writer barbecue we're *always* sending to the same address, and using a connect() call adds some connection verification (and log messages for the user) to help
diagnose misconfiguration (once, during setup, instead of for every single sendto() call).  It also removes the need for tracking the destination address as state (e.g., the socket layer is doing that for us now).  So no, I wouldn't change that.

The biggest problem I see, though, is that setting `self.destination = mc_interface` is going to cause all multicast traffic to be sent only to yourself.  The `mc_interface` variable is for specifying the outgoing ethernet interface (i.e., it's YOUR IP address).  The testcode won't catch this, because it's all running locally, but that one line completely breaks the Multicast writer.

ce-rvi:
Hi @veggiemike ,

Nice to meet you; I'm taking over this conversation on behalf of my colleague @ellapiet who's now on extended leave.

Thanks for your feedback here. Your last point is the salient one, I believe; there was a misinterpretation of the intention of the `mc_interface` kwarg, and this change is intended to make the writer work if the destination multicast address is passed to the writer constructor in `mc_interface` instead of in `destination`. It's probably not worth getting into your other points of discussion until verifying whether the writer works as intended in our environment with the kwargs used as intended, so let me test that and either continue this thread or close it depending on how I go.

Cheers,
Charlie

davidpablocohn:
Thanks for this, Charlie! I hadn't realized that Ella was going to escape quite so quickly for her (much deserved) sabbatical. Please let me know what I can do to help resolve this?

ce-rvi:
EDIT: Realised after the fact I don't have permissions to close the PR, could you give me a hand by doing that please @davidpablocohn? Thanks in advance.

Hi @davidpablocohn (and @veggiemike), 

I've tested with the corrected kwarg usage on our end and all looks well, so I'll close this PR off. Cheers for the discussion both!

Charlie

davidpablocohn:
As per discussion, we're closing this one without pulling

PR #552, opened 2026-05-04
Update default UDP reader interface setup (INADDR_ANY), clarify bind and sockopts logic

ce-rvi:
Hi folks,

Opening this PR with some suggested changes to the UDP reader, see details below. Feedback welcomed, cheers!

- This PR changes how the UDP reader selects a default interface for reading multicast traffic where one is not specified to the constructor.
  - Currently `self.interface = socket.gethostbyname(socket.gethostname())` is used to derive a default value for the interface. This will resolve to `127.0.1.1` on many Linux machines which is not a useable interface for reading UDP multicast traffic. 
  - The proposed alternative is to use socket.INADDR_ANY, which will let the OS pick a sensible default. This is the approach we currently use on RV Investigator (but would like to realign our reader implementation with what's upstream).

Additionally, this PR makes some minor changes in the interest of clarity:
  - Changes the socket bind for multicast from `sock.bind((self.mc_group, self.port))` to `sock.bind(('', self.port))`. To my understanding there is no need to specify the group in the bind once the IGMP group has been joined with the preceding logic. 
  - Removes the logic setting the `socket.IP_MULTICAST_IF` sockopt for multicast. This option only controls which interface is used for multicast send, and this class does not need to send.

### DIFF
--- a/logger/readers/udp_reader.py
+++ b/logger/readers/udp_reader.py
@@ -2,11 +2,7 @@
 
 import logging
 import socket
-<<<<<<< HEAD
-=======
-import sys
 import struct
->>>>>>> cd2d034 (Move to using socket.IPADDR_ANY as the fallback interface for multicast read, and remove redundant IP_MULTICAST_IF sockopt)
 
 from logger.readers.reader import Reader  # noqa: E402
 
--- a/logger/readers/udp_reader.py
+++ b/logger/readers/udp_reader.py
@@ -2,6 +2,11 @@
 
 import logging
 import socket
+<<<<<<< HEAD
+=======
+import sys
+import struct
+>>>>>>> cd2d034 (Move to using socket.IPADDR_ANY as the fallback interface for multicast read, and remove redundant IP_MULTICAST_IF sockopt)
 
 from logger.readers.reader import Reader  # noqa: E402
 
@@ -93,14 +98,6 @@ class UDPReader(Reader):
         if mc_group:
             # resolve once in constructor
             mc_group = socket.gethostbyname(mc_group)
-            if not interface:
-                # multicast needs to specify interface to use, so let's pick a
-                # sane default
-                #
-                # NOTE: This means hostname cannot be an alias to localhost, or
-                #       you won't be able to send IGMP packets correctly.
-                #
-                self.interface = socket.gethostbyname(socket.gethostname())
 
         self.mc_group = mc_group
 
@@ -154,8 +151,6 @@ class UDPReader(Reader):
                                 "specify the interface to use by passing its IP address as the "
                                 "`interface` parameter.  (You can ignore this message if you're "
                                 "actually just doing loopback testing.)")
-            sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_IF,
-                            socket.inet_aton(self.interface))
 
             # join the group via IGMP
             #
@@ -164,11 +159,21 @@ class UDPReader(Reader):
             #       could use struct.pack("4s4s", ...) to create a struct to
             #       pass into setsockopt()
             #
-            sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
-                            socket.inet_aton(self.mc_group) + socket.inet_aton(self.interface))
+            if self.interface:
+                sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
+                                socket.inet_aton(self.mc_group) + socket.inet_aton(self.interface))
+            else:
+                mreq = struct.pack("4sl", socket.inet_aton(self.mc_group), socket.INADDR_ANY)
+                sock.setsockopt(
+                    socket.IPPROTO_IP, 
+                    socket.IP_ADD_MEMBERSHIP, 
+                    mreq
+                )
 
-            # bind to mc_group:port
-            sock.bind((self.mc_group, self.port))
+
+            # bind
+            # don't need to specify mc_group address because we have already joined above
+            sock.bind(('', self.port))
 
         else:
             # broadcast or unicast, bind to specificed interface
--- a/logger/readers/udp_reader.py
+++ b/logger/readers/udp_reader.py
@@ -44,9 +44,7 @@ class UDPReader(Reader):
         """
         ```
         interface    IP (or resolvable name) of interface to listen on.  None or ''
-                     will listen on INADDR_ANY (all interfaces).  If joining a
-                     multicast group and None or '' specified, this will default
-                     to whatever the system's hostname resolves to.  This IP should
+                     will listen on INADDR_ANY (all interfaces). This IP should
                      not be on the loopback network (OK for testing, but won't work
                      in the real world).
 
@@ -70,13 +68,6 @@ class UDPReader(Reader):
         """
         super().__init__(**kwargs)
 
-        if interface:
-            # resolve once in constructor
-            interface = socket.gethostbyname(interface)
-        else:
-            interface = ''
-        self.interface = interface
-
         # make sure user passed in `port`
         #
         # NOTE: We want the order of the arguments to consistently be (ip,
@@ -90,12 +81,17 @@ class UDPReader(Reader):
         # make sure port gets stored as an int, even if passed in as a string
         self.port = int(port)
 
-        # prep multicast parameters
+        # resolve mc_group once in constructor
         if mc_group:
-            # resolve once in constructor
-            mc_group = socket.gethostbyname(mc_group)
-
+            mc_group = socket.inet_aton(socket.gethostbyname(mc_group))
         self.mc_group = mc_group
+
+        # resolve interface once in constructor
+        if interface:
+            interface = socket.inet_aton(socket.gethostbyname(interface))
+        else:
+            interface = socket.INADDR_ANY if self.mc_group else socket.inet_aton('0.0.0.0')
+        self.interface = interface
 
         self.reuseaddr = reuseaddr
         self.reuseport = reuseport
@@ -140,7 +136,7 @@ class UDPReader(Reader):
             #       never leave the system, and you never actually join the
             #       group.
             #
-            if self.interface.startswith('127.') and not self.this_is_a_test:
+            if socket.inet_ntoa(self.interface).startswith('127.') and not self.this_is_a_test:
                 logging.warning("Can't use loopback device for joining multicast groups.  Make "
                                 "sure your system's hostname does NOT resolve to something in "
                                 "the 127.0.0.0/8 address block (e.g., localhost, 127.0.0.1), or "
@@ -149,31 +145,15 @@ class UDPReader(Reader):
                                 "actually just doing loopback testing.)")
 
             # join the group via IGMP
-            #
-            # NOTE: Since these are both already encoded as binary by
-            #       inet_aton(), we can just concatenate them.  Alternatively,
-            #       could use struct.pack("4s4s", ...) to create a struct to
-            #       pass into setsockopt()
-            #
-            if self.interface:
-                sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
-                                socket.inet_aton(self.mc_group) + socket.inet_aton(self.interface))
-            else:
-                mreq = struct.pack("4sl", socket.inet_aton(self.mc_group), socket.INADDR_ANY)
-                sock.setsockopt(
-                    socket.IPPROTO_IP, 
-                    socket.IP_ADD_MEMBERSHIP, 
-                    mreq
-                )
+            sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
+                            self.mc_group + self.interface)
 
-
-            # bind
-            # don't need to specify mc_group address because we have already joined above
-            sock.bind(('', self.port))
+            # multicast, bind to mc_group
+            sock.bind((socket.inet_ntoa(self.mc_group), self.port))
 
         else:
             # broadcast or unicast, bind to specificed interface
-            sock.bind((self.interface, self.port))
+            sock.bind((socket.inet_ntoa(self.interface), self.port))
 
         return sock
 
--- a/logger/readers/udp_reader.py
+++ b/logger/readers/udp_reader.py
@@ -2,7 +2,6 @@
 
 import logging
 import socket
-import struct
 
 from logger.readers.reader import Reader  # noqa: E402
 
@@ -68,6 +67,13 @@ class UDPReader(Reader):
         """
         super().__init__(**kwargs)
 
+        # resolve interface once in constructor
+        if interface:
+            interface = socket.inet_aton(socket.gethostbyname(interface))
+        else:
+            interface = socket.inet_aton('0.0.0.0')
+        self.interface = interface
+
         # make sure user passed in `port`
         #
         # NOTE: We want the order of the arguments to consistently be (ip,
@@ -85,13 +91,6 @@ class UDPReader(Reader):
         if mc_group:
             mc_group = socket.inet_aton(socket.gethostbyname(mc_group))
         self.mc_group = mc_group
-
-        # resolve interface once in constructor
-        if interface:
-            interface = socket.inet_aton(socket.gethostbyname(interface))
-        else:
-            interface = socket.INADDR_ANY if self.mc_group else socket.inet_aton('0.0.0.0')
-        self.interface = interface
 
         self.reuseaddr = reuseaddr
         self.reuseport = reuseport
@@ -145,6 +144,12 @@ class UDPReader(Reader):
                                 "actually just doing loopback testing.)")
 
             # join the group via IGMP
+            #
+            # NOTE: Since these are both already encoded as binary by
+            #       inet_aton(), we can just concatenate them.  Alternatively,
+            #       could use struct.pack("4s4s", ...) to create a struct to
+            #       pass into setsockopt()
+            #
             sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
                             self.mc_group + self.interface)
 


davidpablocohn:
Thanks for this! I'd love to have @veggiemike sanity check this, as he's much closer to where the rubber meets the road with respect to UDP.

logger/readers/udp_reader.py
davidpablocohn: Default is gone, so now need to test that self.interface is not None prior to looking at properties:

```
if self.interface and self.interface.startswith('127.') and not self.this_is_a_test:
```

ce-rvi: Thanks for this David. 

I believe the block at lines 73-78 already protects against this case, since a falsy value of `interface` will lead to `self.interface` being set to `''`. The logic I am proposing to remove from the constructor was then overwriting the value of `self.interface` for multicast contexts. But certainly would be happy to add this to make it more explicit.

ce-rvi:
> Thanks for this! I'd love to have @veggiemike sanity check this, as he's much closer to where the rubber meets the road with respect to UDP.

Thanks David. I'd definitely appreciate @veggiemike 's perspective as there was obviously a reasoning behind defaulting the interface in this way for multicast which would be great to understand.

logger/readers/udp_reader.py
davidpablocohn: Ah - you're right! Sorry about that. Disregard...

veggiemike:
FYI, I'm planning to review this sometime this week... just struggling to find the time.  ;-)

veggiemike:
> > Thanks for this! I'd love to have @veggiemike sanity check this, as he's much closer to where the rubber meets the road with respect to UDP.
> 
> Thanks David. I'd definitely appreciate @veggiemike 's perspective as there was obviously a reasoning behind defaulting the interface in this way for multicast which would be great to understand.

This is traditionally how I've done it on other projects, in an attempt to guess a "correct" multicast interface to use for the IGMP requests, hopefully limiting what has to get set in a config file.  Since multicast traffic doesn't follow normal routing principles, you have to explicitly specify which ethernet interface to use when joining the IGMP group.  Doesn't matter when you only have one, but does when you have multiple (at least, it did when I tested this thoroughly years ago).

Could alternatively default to ~~localhost if unspecified, which is completely invalid for anything other than basic testing, and force users to specify their outgoing multicast interface all the time~~ `INADDR_ANY` as you're suggesting and let the OS choose, but this "resolve your hostname" thing proved to be correct most of the time on the systems I was working on.

(Not sure why my brain jumped off the track and started talking about localhost up there... ignore that)

logger/readers/udp_reader.py
```diff
@@ -154,8 +151,6 @@ def _open_socket(self):
                                 "specify the interface to use by passing its IP address as the "
                                 "`interface` parameter.  (You can ignore this message if you're "
                                 "actually just doing loopback testing.)")
-            sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_IF,
-                            socket.inet_aton(self.interface))
```
veggiemike, line 158: We really *do* need set the multicast interface.  Otherwise people running this on multi-homed machines won't be able to specify which interface to receive multicast traffic on.

veggiemike: Actually, now that I've said that, I think you're right...  you only have to do the IP_MULTICAST_IF call if you're sending multicast traffic.  The 2nd param to the IP_ADD_MEMBERSHIP call below is how we specify the interface to receive on.

logger/readers/udp_reader.py
```diff
@@ -164,11 +159,21 @@ def _open_socket(self):
...
+            else:
+                mreq = struct.pack("4sl", socket.inet_aton(self.mc_group), socket.INADDR_ANY)
+                sock.setsockopt(
+                    socket.IPPROTO_IP, 
+                    socket.IP_ADD_MEMBERSHIP, 
+                    mreq
+                )
+
 
-            # bind to mc_group:port
```
veggiemike, line 170: The intent here, even though you don't *have* to bind to the mc_group IP, was to hopefully have `netstat -pnave --udp` output display the mc_group IP instead of INADDR_ANY.  That being said, I don't recall if it actually had that effect.  It does work, though, and is simply more explicit than binding to INADDR_ANY.

logger/readers/udp_reader.py
```diff
@@ -93,14 +98,6 @@ def __init__(self, interface=None, port=None, mc_group=None,
...
             # resolve once in constructor
             mc_group = socket.gethostbyname(mc_group)
-            if not interface:
-                # multicast needs to specify interface to use, so let's pick a
-                # sane default
-                #
-                # NOTE: This means hostname cannot be an alias to localhost, or
-                #       you won't be able to send IGMP packets correctly.
-                #
-                self.interface = socket.gethostbyname(socket.gethostname())
```
veggiemike, line 103: The way this was written, we *always* set the multicast interface using self.interface, but we try to guess a sane default if it wasn't specified.  I was trying to avoid using INADDR_ANY by default for this because if your system has multiple ethernet interfaces, the OS will "pick one", the behavior of which is undefined as far as I know, almost definitely varies by OS, and probably not what you want at least half the time.

If what you'd like to see happen is let the OS pick if we haven't specified (which is admittedly probably more likely to be expected by developers), instead of making this resolv-your-hostname guess, then I'd say we should simply set `self.interface = socket.INADDR_ANY` right here and leave the existing IP_ADD_MEMBERSHIP setsockopt code below alone.

But then, if we're changing the class's behavior, you have to update the docstring at the top of __init__ as well.

logger/readers/udp_reader.py
```diff
@@ -164,11 +159,21 @@ def _open_socket(self):
             #       could use struct.pack("4s4s", ...) to create a struct to
             #       pass into setsockopt()
             #
-            sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP,
-                            socket.inet_aton(self.mc_group) + socket.inet_aton(self.interface))
+            if self.interface:
```
veggiemike: As suggested above, just blindly use `self.interface` as originally written w/out any conditional code.  We've either been passed in a value or already assigned it to `INADDR_ANY`.

veggiemike:
And lastly, apologies for taking so long to look at this.  Saying "I've been busy" would be a gross understatement.  :rofl:

logger/readers/udp_reader.py
ce-rvi: That's my understanding as well -- IP_MULTICAST_IF for sending, IP_ADD_MEMBERSHIP for receiving 👍

logger/readers/udp_reader.py
ce-rvi: I'm totally happy in principle to avoid having conditional code here, however it's a little fiddlier than just setting `self.interface = socket.INADDR_ANY` in the constructor, because the existing code here calls `inet_aton` on the value, which we can't pass `socket.INADDR_ANY` to. We could set `self.interface = socket.inet_aton(interface)` _or_  `socket.INADDR_ANY` in the constructor to avoid conditional logic here.

An advantage of the conditional code is that it leaves your battle-tested implementation in place for the one path, and uses our Investigator-battle-tested implementation for the other path.

No strong feelings about this one though, happy to do what the crowd is happy with :)

logger/readers/udp_reader.py
ce-rvi: Fair enough, happy to roll this back 👍

logger/readers/udp_reader.py
ce-rvi: This is interesting and I'd like to understand better how our environments differ, because at least in my environment, I would never expect the current implementation to provide a sane default, but would usually expect the proposed implementation to.

From my side, we are deploying on an Ubuntu server. [As you would expect for a 'normal' Ubuntu / Debian server](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution), there is an entry in `/etc/hosts` like `127.0.1.1 my-example-hostname`. As our `/etc/nsswitch.conf` configuration is also left to defaults, `/etc/hosts` will be preferred over DNS as a source for host resolution. So, if I call `socket.gethostbyname(socket.gethostname())` on my system, I am going to be returned `127.0.1.1`, which is not a usable interface for multicast because it's in the loopback range. 

In contrast, notwithstanding that the OS might not pick the right or best interface if I specify socket.INADDR_ANY, it should at least pick a non-loopback interface, and (I won't pretend to have a great understanding of what's happening under the hood, but) an interface that it expects to be able to receive on from the specified multicast group.

Curious how this reads from your side, is there anything obviously different in our environments that would account for our different experiences?

ce-rvi:
Hi @veggiemike, no worries and thanks for finding the time!

I've replied to your comments individually.

logger/readers/udp_reader.py
veggiemike: > This is interesting and I'd like to understand better how our environments differ, because at least in my environment, I would never expect the current implementation to provide a sane default, but would usually expect the proposed implementation to.
>
> From my side, we are deploying on an Ubuntu server. [As you would expect for a 'normal' Ubuntu / Debian server](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution), there is an entry in /etc/hosts like 127.0.1.1 my-example-hostname. As our /etc/nsswitch.conf configuration is also left to defaults, /etc/hosts will be preferred over DNS as a source for host resolution. So, if I call socket.gethostbyname(socket.gethostname()) on my system, I am going to be returned 127.0.1.1, which is not a usable interface for multicast because it's in the loopback range.

This is practically a religious debate among old-school Linux users.  ;-)  Traditionally, RHEL does NOT set the hostname to an alias in `/ec/hosts` because servers should be able to resolve their IPs (and it's actually considered INCORECT for your hostname to resolve to localhost).  In contrast, Fedora sets a localhost-ish alias to the hostname in `/etc/hosts` just like Ubuntu now does, because these distros cater to home/mobile users.  Ubuntu Server does the same thing as regular Ubuntu (pretty sure), but then they're practically the same product compared to the RHEL/Fedora split.

So, in a RHEL environment (at least as of a few years ago), it's perfectly safe to assume your hostname resolves to your actual IP address.  I primarily use Ubuntu (now, because new job was all Debian/Ubuntu), but can't stand the localhost alias and remove it on my servers (I've run down multiple troubleshooting rabbit holes over the years that went off the rails because of bad assumptions being made because the hostname is resolving to localhost).  I even go so far as to patch the OpenRVDAS install script to NOT make that alias (pretty sure it still does that).

That being said, years removed from writing this code, I agree you're right: the use case for my resolve-hostname-to-guess is overcomplicated and is going to definitely be wrong on a non-trivial percentage of distros.  Just using INADDR_ANY will at least be correct all the time on systems with only a single NIC.  Defaulting to INADDR_ANY seems much more "general case".

> In contrast, notwithstanding that the OS might not pick the right or best interface if I specify socket.INADDR_ANY, it should at least pick a non-loopback interface, and (I won't pretend to have a great understanding of what's happening under the hood, but) an interface that it expects to be able to receive on from the specified multicast group.

The problem is that in an environment with multiple network interfaces, the kernel CANNOT make a smart choice.  Multicast traffic doesn't follow normal routing rules, so it's going to do something silly but hopefully at least deterministic (e.g., always pick the interface with the lowest MAC address).  So if you have 2 interfaces, say eth0 on 10.1.10.0/24 and eth1 on 192.168.1.0/24, and INADDR_ANY happens to cause eth1 to be used, your IGMP messages go to 192.168.1.255.  But if the actual multicast server is sitting on 10.1.10.0/24, it won't get your IGMP message at all... and you'll get no multicast traffic even though you "joined the group".

logger/readers/udp_reader.py
veggiemike: > I'm totally happy in principle to avoid having conditional code here, however it's a little fiddlier than just setting self.interface = socket.INADDR_ANY in the constructor, because the existing code here calls inet_aton on the value, which we can't pass socket.INADDR_ANY to. We could set self.interface = socket.inet_aton(interface) or socket.INADDR_ANY in the constructor to avoid conditional logic here.

Yep, I missed that.  I still like setting it while checking params at the top of the function and then just using it, just make sure it's the same data type.  So something like this:
```Python
        if interface:
            # resolve once in constructor                                                                                               
            self.interface = socket.inet_aton(socket.gethostbyname(interface))
        else:
            self.interface = socket.INADDR_ANY
```
Then further down we can just use `self.interface` consistently w/out calling `inet_aton`.

While we're at it, we should also prep `self.mc_group` in the constructor if specified (i.e., make sure if `inet_aton` is gonna fail, it happens during construction).  So stick something like this right after we set `self.interface`:
```Python
        if mc_group:
            # resolve once in constructor
            self.mc_group = socket.inet_aton(socket.gethostbyname(mc_group))
        else:
            self.mc_group = None
```
That way both `self.interface` and `self.mc_group` are ready to get passed into `getsockopt` calls after construction.

veggiemike:
Cheers from somewhere near Hawaii!

logger/readers/udp_reader.py
ce-rvi: Thanks for the explanation @veggiemike , I've never worked with RHEL so that's news to me but useful to know! And point well taken RE: kernel not being able to make a smart choice between multiple NICs.

Notwithstanding, I'm taking away that you're happy to adopt `INADDR_ANY` as the default case all things considered (correct me if I'm wrong)

logger/readers/udp_reader.py
ce-rvi: Agree with what you've suggested here, thanks -- slightly under the pump with other work but will aim to push a commit soon!

veggiemike: No problem, I know the feeling.

logger/readers/udp_reader.py
veggiemike: Absolutely.  :sunglasses:

ce-rvi:
@veggiemike apols again for the delay, I've been pulled every which way this month! Would you mind taking another look at this in light of my latest commit? I've moved all the conditional `mc_group` / `interface` logic up into the constructor as discussed which I think works nicely. It did have the unintended consequence of complicating the `sock.bind` calls at the bottom of `_open_socket` because  as far as I know, `bind` only takes the address in string form ... not a showstopper, I've just converted them back as required with `socket.inet_ntoa`, but let me know if any objections. Similarly, I had to change the default unicast 'interface' value from `''` to `socket.inet_aton('0.0.0.0')` to make `self.interface` type-consistent across the board, but this is functionally equivalent to my understanding.

Thanks in advance if you've time to take a look!

logger/readers/udp_reader.py
```diff
@@ -89,21 +81,18 @@ def __init__(self, interface=None, port=None, mc_group=None,
...
-                self.interface = socket.gethostbyname(socket.gethostname())
-
+            mc_group = socket.inet_aton(socket.gethostbyname(mc_group))
         self.mc_group = mc_group
 
+        # resolve interface once in constructor
+        if interface:
+            interface = socket.inet_aton(socket.gethostbyname(interface))
+        else:
+            interface = socket.INADDR_ANY if self.mc_group else socket.inet_aton('0.0.0.0')
```
veggiemike:
```suggestion
            interface = socket.inet_aton('0.0.0.0')
```
You're setting `interface` to the integer 0 (aka `socket.INADDR_ANY`) for multicast... and then passing that into `inet_ntoa()` to check for localhost usage, which will go boom.  Just set it to `socket.inet_aton('0.0.0.0')` universally.

As a general note, when doing socket programming in Python, you will almost never use socket.INADDR_ANY and the other socket.INADDR_* constants...  They're pulled out of the C library as integers, but all the python functions that *would* use them only take binary string objects (e.g., `b'\x00\x00\x00\x00'` aka the return value of `socket.inet_aton('0.0.0.0')`)

logger/readers/udp_reader.py
```diff
@@ -2,6 +2,7 @@
 
 import logging
 import socket
+import struct
```
veggiemike:
```suggestion
```
We don't need this anymore, the struct module isn't used anywhere in here.

logger/readers/udp_reader.py
```diff
@@ -147,32 +136,24 @@ def _open_socket(self):
...
-            sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_IF,
-                            socket.inet_aton(self.interface))
 
             # join the group via IGMP
-            #
-            # NOTE: Since these are both already encoded as binary by
-            #       inet_aton(), we can just concatenate them.  Alternatively,
-            #       could use struct.pack("4s4s", ...) to create a struct to
-            #       pass into setsockopt()
-            #
```
veggiemike, line 166: I'd actually put this comment block back.  It's still perfectly valid (we do still pass both values through `inet_aton()`, just in the constructor now), and points out why we're doing things that might look "wrong" when compared to example code.

logger/readers/udp_reader.py
```diff
@@ -69,13 +68,6 @@ def __init__(self, interface=None, port=None, mc_group=None,
         """
         super().__init__(**kwargs)
 
-        if interface:
-            # resolve once in constructor
-            interface = socket.gethostbyname(interface)
-        else:
-            interface = ''
-        self.interface = interface
-
```
veggiemike, line 78: Can we move the new block of code for handling `interface` up here so the diffs read nicer?

veggiemike (changes requested):
Looks pretty good.  1 actual issue, and a couple trivial fixups.

ce-rvi:
Cheers for the latest review @veggiemike , have just implemented your suggestions. Thanks for the catch on the 'boom'.

davidpablocohn:
So we're good to go?

veggiemike:
> So we're good to go?

I'm happy with the code changes.  I was planning to actually do a smidge of testing before officially signing off... and then life got hectic.  I can probably get to it this afternoon.

veggiemike:
@davidpablocohn this is good to merge.  I didn't thoroughly test the multicast code (I don't have a test setup for that at the moment), but did verify no obvious breakage in the regular unicast UDP Reader code.

veggiemike (approved):
Looks good!

PR #560, opened 2026-05-25
Issue #503: Add React/FastAPI web UI

webbpinner:
## Summary

- Adds a React 19 + Vite + TypeScript frontend (`web_frontend`) and a FastAPI + SQLAlchemy backend (`web_backend`) as git submodules, providing a modern web UI alternative to the existing Django UI
- Integrates with OpenRVDAS via the CachedDataServer and logger_manager, supporting cruise/logger/mode management, live status, configuration preview/apply, connection testing, and parser verification
- Includes install script support: the existing `install_openrvdas.sh` prompts for React UI installation, sets up the FastAPI backend venv, configures `.env` files, and registers services with supervisord

## What's in this PR

**Backend (`web_backend` → `e654399`)**
- JWT + API key authentication with role-based access
- Proxy routes for CachedDataServer data, loggers, cruise, modes, and updates
- Configuration file browser, preview/validate, and apply endpoints
- WebSocket and SSE test-connection endpoint; parser format verification
- SQLite-backed (no PostgreSQL dependency on this branch)
- 61-test pytest suite; pre-commit hooks (black, isort, flake8, mypy)

**Frontend (`web_frontend` → `1d2d742`)**
- Dashboard with live logger status and log viewer
- Cruise configuration load/preview/apply flow with two-step validation
- Template builder and verify-parser pages
- WebSocket-streaming test-connection page
- Fix: `npm run build` / `type-check` hanging on TypeScript 5.8 (tsc -b + noEmit regression)
- Fix: `VITE_OPENRVDAS_VERSION` now read from `.env` rather than `../pyproject.toml` at build time

## Test plan

- [x] Run `install_openrvdas.sh` on a clean Ubuntu 24.04 VM, selecting React UI option
- [ ] Verify backend starts and `/api/v1/docs` is accessible
- [ ] Verify frontend builds and serves correctly
- [ ] Log in, load a cruise configuration, confirm loggers appear and modes switch
- [ ] Test live logger status updates via WebSocket
- [ ] Test connection testing and parser verification pages
- [ ] Confirm `npm run build` completes without hanging


davidpablocohn:
Hmm - worked on MacOS for me, but on a fresh Ubuntu 24 install with react front end, system won't let me log in. Claude believes:

_The frontend was built with VITE_SERVER_API_BASE_URL=http://localhost:8000 baked in — so it's bypassing nginx entirely and hitting the backend directly from the browser
  (which fails due to CORS + wrong host).

The real fix is to change the Zod default in config.ts from "http://localhost:8000" to "" and handle an empty string as
  window.location.origin in apiUrl()._

Recommendation:
  1. nginx not configured — openrvdas_nginx.conf exists in web_frontend/ but the install script never copies it to /etc/nginx/sites-enabled/. Also the file uses macOS paths
   (/opt/homebrew/...) so it needs to be Linux-adapted.
  2. VITE_SERVER_API_BASE_URL defaults to http://localhost:8000 — when blank/unset it should default to window.location.origin (so the frontend goes through nginx), but the
   Zod default in config.ts hardcodes localhost instead.
  3. logger_manager needs PYTHONPATH=/opt/openrvdas in its supervisor config environment, or the install script needs to ensure it's set.

webbpinner:
Claude was right... applied the fix.  Please give it another try.

davidpablocohn:
Excellent - thanks!

Two minor issues remaining - neither are showstoppers, but as long as we're at it. Claude has figured out why sometimes a reboot is needed before everything works:

  PR #560 Review — Round 2 (2026-05-26)
  
  Tested on a fresh Ubuntu install at https://openrvdas-test. Login, dashboard, logger status, and cruise loading all work after working through the issues below.

  ---
  Issue 1: PYTHONPATH not set for logger_manager and cached_data_server (Bug)
  
  The supervisord config generated by the install script sets directory=/opt/openrvdas but does not set PYTHONPATH. When supervisor runs python server/logger_manager.py,
  Python adds /opt/openrvdas/server/ to sys.path (the script's directory), not /opt/openrvdas/. So from server.cached_data_server import ... fails with ModuleNotFoundError:
   No module named 'server' on any fresh start.

  Fix — in the install script, change the environment= line for both logger_manager and cached_data_server:
  environment=PYTHONPATH="${INSTALL_ROOT}/openrvdas",PATH="${VENV_BIN}:/usr/bin:/usr/local/bin"
  
  ---
  Issue 2: System nginx conflicts with supervisord-managed nginx (Installation bug)
  
  The install script correctly stops and disables the system nginx service, and supervisord runs nginx directly with the generated full config. However, on a machine where
  the system nginx was already running with a default site in /etc/nginx/sites-enabled/, the supervisord nginx fails to bind port 80 ("duplicate default server" error). The
   install script should explicitly remove /etc/nginx/sites-enabled/default (or all sites-enabled entries) as part of setup_nginx() before writing the new config.

  ---
  Issue 3: Startup ordering — web console doesn't show logger state until reboot (UX bug)
  
  On first install, the web console showed loggers as not running even though they were. A reboot fixed it. The cause is that uvicorn starts before logger_manager and
  cached_data_server are fully up, so the initial status snapshot is empty and no update notification fires to trigger a refresh.

  Fix — add priority= to the supervisord configs to enforce startup order:
  - cached_data_server: priority=10
  - logger_manager: priority=20
  - react_ui group (nginx + uvicorn): priority=30

  ---
  Items confirmed fixed from round 1 review:
  
  - VITE_SERVER_API_BASE_URL now defaults to "" with window.location.origin fallback — frontend correctly routes through nginx proxy
  - nginx config now uses OS-aware mime.types path (no more hardcoded /opt/homebrew/...)
  - All 8 bugs from the original code review (path traversal, logger.running, get_event_loop, get_status shape, debug prints, WS refresh token bypass, dead hook) confirmed
  fixed

davidpablocohn:
Sorry - found one more Ubuntu-related issue: looks like there's a second place the VITE_SERVER_API_BASE_URL needs to be updated? (change .env.dist).

Issue 4: .env.dist hardcodes localhost:8000, breaking every fresh install (Bug)
  
  web_frontend/.env.dist contains:
  # URL of the backend API. Leave unset to default to window.location.origin.
  VITE_SERVER_API_BASE_URL=http://localhost:8000
  
  The install script copies .env.dist to .env on a fresh install, so Vite bakes http://localhost:8000 into the JS bundle. The browser then makes direct requests to
  localhost instead of going through nginx, causing CORS errors on every fresh install. The comment on the line even says "Leave unset to default to window.location.origin"
   — the value contradicts it.

  The config.ts fix (defaulting VITE_SERVER_API_BASE_URL to "" with window.location.origin fallback) was correct but incomplete. Fix: change .env.dist to:
  VITE_SERVER_API_BASE_URL=

davidpablocohn:
One last bit remaining! I failed to catch this and add it to previous:

Issue 4: .env.dist hardcodes localhost:8000, breaking every fresh install (one small thing remaining)
  
  This is the last piece — almost there! web_frontend/.env.dist still contains:
  # URL of the backend API. Leave unset to default to window.location.origin.
  VITE_SERVER_API_BASE_URL=http://localhost:8000
  
  The install script copies .env.dist to .env on a fresh install, so Vite bakes http://localhost:8000 into the JS bundle. The browser then makes direct requests to
  localhost instead of going through nginx, causing CORS errors. This happens even though the config.ts fix (defaulting to "" with window.location.origin fallback) is
  correct — .env.dist overrides it at build time.

  The fix is a one-line change in the web_frontend submodule:
  VITE_SERVER_API_BASE_URL=
  
  The comment already says "Leave unset to default to window.location.origin" — this just makes the value match the comment.

I've tried patching on the test machine with:

  sed -i 's|^VITE_SERVER_API_BASE_URL=.*|VITE_SERVER_API_BASE_URL=|' /opt/openrvdas/web_frontend/.env.dist
  sed -i 's|^VITE_SERVER_API_BASE_URL=.*|VITE_SERVER_API_BASE_URL=|' /opt/openrvdas/web_frontend/.env
  cd /opt/openrvdas/web_frontend && npm run build
  sudo supervisorctl restart react_ui:nginx

and verified that it worked.

webbpinner:
ugh... I knew about that and forgot to correct it... one sec.

davidpablocohn:
Thank you! All that seems to work now.  Aaaand I've tickled one minor startup bug - after installation and initial startup, the console doesn't register that the loggers are actually running. A supervisorctl restart all fixes.

Sorry - no good deed goes unpunished, I know, I know.

Claude flags the problem as this:

  Issue 5: _db_poll_loop never sends an initial update on WebSocket connect (root cause of Issue 3)
  
  After a fresh boot, the console still shows all loggers as "exited" until supervisorctl restart all. The priority ordering from Issue 3 helps but doesn't fully cover this
   case.

  The root cause is in web_backend/app/api/updates.py, _db_poll_loop():
```
  if last_ts is None:
      last_ts = ts          # just snapshots current timestamp — never sends
  elif ts != last_ts:
      last_ts = ts
      await websocket.send_json({"type": "update"})
  ```
  On first iteration, the loop snapshots the current LastUpdate timestamp and waits for it to change. If logger_manager finished writing its startup state before the
  frontend's WebSocket connected (which the priority ordering makes more likely, not less), the timestamp never changes and the frontend never gets a refresh — it stays on
  whatever stale state it loaded initially.
  
  After supervisorctl restart all, logger_manager writes to the DB while the frontend is already connected, the timestamp changes, and the update fires correctly — which is
   why that workaround works.

  The fix is one line:
```
  if last_ts is None:
      last_ts = ts
      await websocket.send_json({"type": "update"})  # always sync fresh state on connect
```
  This guarantees the frontend fetches current state as soon as the WebSocket connects, regardless of boot order.

davidpablocohn (approved):

PR #576, opened 2026-06-22
Issue 574 - Fix log_entries not reaching React UI

webbpinner:
## Summary

Fixes four root causes preventing log entries from appearing in the React UI on RHEL/AlmaLinux installations:

- WebSocket frame size limit (web_backend/app/api/updates.py): The websockets library default 1 MB max_size was silently killing _cds_log_loop on its first connect. A backfill of 100 log records across many loggers easily exceeds this limit — the CDS closed the connection with 1009 message too big, first_connect was already False, so all retries used seconds=0 and never attempted the backfill again. Fixed by setting max_size=None on both websockets.connect() calls.
- SELinux blocking CDS connections (utils/install_openrvdas.sh): On RHEL/AlmaLinux, httpd_can_network_connect was off by default and port 8766 (CDS) was unlabeled, preventing uWSGI/FastAPI from reaching the CachedDataServer. Extracted SELinux configuration into a dedicated setup_selinux() function that runs unconditionally on CentOS/RHEL, independent of the firewall setup choice.
- _db_poll_loop missing exception guard (web_backend/app/api/updates.py): An unhandled DB exception in _db_poll_loop could propagate through asyncio.wait(FIRST_COMPLETED) and cancel the sibling _cds_log_loop task. Wrapped the DB query in a try/except.
- wsBaseUrl() missing fallback (web_frontend/src/hooks/useLoggerStateWS.ts): When VITE_SERVER_API_BASE_URL is unset (the default for same-origin production builds), wsBaseUrl() returned an empty string, producing an invalid WebSocket URL. Now falls back to window.location.origin, mirroring the pattern already used in api.tsx.

## Tests

- Verified log entries appear in the React UI on an AlmaLinux 9 install after pulling and restarting the FastAPI service.

Related: https://github.com/OceanDataTools/openrvdas/issues/574

### DIFF
--- a/utils/install_openrvdas.sh
+++ b/utils/install_openrvdas.sh
@@ -1657,6 +1657,36 @@ EOF
 
 ###########################################################################
 ###########################################################################
+# CentOS/RHEL ONLY - Configure SELinux for OpenRVDAS services
+function setup_selinux {
+    if [ $OS_TYPE == 'MacOS' ] || [ $OS_TYPE == 'Ubuntu' ]; then
+        return
+    fi
+
+    echo "#####################################################################"
+    echo "Configuring SELinux permissions for OpenRVDAS"
+
+    # The old way of enabling things...
+    # (sed -i -e 's/SELINUX=enforcing/SELINUX=permissive/g' /etc/selinux/config) || echo UNABLE TO UPDATE SELINUX! Continuing...
+
+    # Allow the web server process to make outbound network connections (needed
+    # for uWSGI/FastAPI to reach the CachedDataServer and other services).
+    setsebool -P nis_enabled 1
+    setsebool -P use_nfs_home_dirs 1
+    setsebool -P httpd_can_network_connect 1
+    semanage permissive -a httpd_t
+
+    # Label the CachedDataServer's fixed port so httpd_t can connect to it.
+    # Port 8766 is the universal CDS port across all OpenRVDAS installations.
+    semanage port -a -t http_port_t -p tcp 8766 2>/dev/null || \
+        semanage port -m -t http_port_t -p tcp 8766
+
+    echo "Done configuring SELinux permissions"
+}
+
+
+###########################################################################
+###########################################################################
 # CentOS/RHEL ONLY - Set up firewall daemon and open relevant ports
 function setup_firewall {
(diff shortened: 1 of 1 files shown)


web_backend
```diff
@@ -1 +1 @@
-Subproject commit fe216e5bc1e44f2984e8f7179942ecf9a353d720
+Subproject commit 15d425982807c5eb72fda5b8ceff1b13d9bef25f
```
davidpablocohn: **Re: `web_backend/app/api/updates.py` `_db_poll_loop` (lines ~37–55 of this submodule bump)**

Posting here since the actual code lives in the `fastapi-template` submodule and isn't in this PR's diff (hence a plain block rather than an applyable suggestion).

Minor, non-blocking: the new `except Exception: pass` silently swallows *all* errors, including a genuine DB failure (which then spins every 0.5s with no trace) and the client `send_json`. Scoping the `try` to just the DB query — and logging it like the `_cds_*` loops do — keeps real failures visible and mirrors the existing `return  # Client disconnected` idiom for the client send. Behavior is otherwise identical (first poll always sends, then only on change):

```python
    while True:
        try:
            async with AsyncSessionLocal() as session:
                result = await session.execute(
                    select(LastUpdate).order_by(LastUpdate.timestamp.desc()).limit(1)
                )
                latest = result.scalar_one_or_none()
                ts = latest.timestamp if latest else None
        except Exception as e:
            logger.debug("DB poll loop: %s: %s", type(e).__name__, e)
            await asyncio.sleep(0.5)
            continue

        if last_ts is None or ts != last_ts:
            last_ts = ts
            try:
                await websocket.send_json({"type": "update"})
            except Exception:
                return  # Client disconnected

        await asyncio.sleep(0.5)
```

Functionally the current version is safe — both sibling loops `return` on client disconnect so teardown still fires via `FIRST_COMPLETED` — so treat this purely as a readability/observability nit.

webbpinner: Applied — pushed `2e35d94` to fastapi-template's `openrvdas` branch, scoping the try/except to just the DB query, logging it via `logger.debug`, and returning on `send_json` failure to distinguish client-disconnect from a real DB error. Submodule pointer bumped here in `7b5b9a8f`.

PR #577, opened 2026-06-22
MQTTReader: make client_name optional and ensure client_id is unique

veggiemike:
This commit makes the `client_name` constructor param optional, allowing the underlying Paho MQTT library to generate a random ID as needed.  If `client_name` is provided, it is appended with 6 random characters to help ensure uniqueness.

W/out this patch, if you accidentally supply the same `client_name` for multiple readers (say, via a template), each new connection causes the broker to disconnect all previous instances, and only the last instance receives messages.

### DIFF
--- a/logger/readers/mqtt_reader.py
+++ b/logger/readers/mqtt_reader.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 import logging
+import random
+import string
 from queue import Queue
 
 # Don't barf if they don't have redis installed. Only complain if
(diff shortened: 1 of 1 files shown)


davidpablocohn (approved):

davidpablocohn:
Ah - just realized we merged this into master. In general I like to stage by merging into dev first. I'll go ahead and cherry pick the change, then merge it into dev as well.

veggiemike:
> Ah - just realized we merged this into master. In general I like to stage by merging into dev first. I'll go ahead and cherry pick the change, then merge it into dev as well.

Oops, that's my bad.  I didn't even think about it when I wrote the pull request.

PR #588, opened 2026-08-17
Fix typo in logging message in udp_reader.py

peter-csiro:
Noticed tiny typo while reading the code. Not urgent or anything as it's just a log message, but hopefully also a trivial thing to fix with no side effects (assuming no one is depending on fixed string log messages :)).


davidpablocohn:
Oops - thank you! I'm going to sit on this for a couple of days to see what other bugs pop up with 2.6.0 and try to pull them all in together.

davidpablocohn (approved):
LGTM - thanks!

Issue #621, opened 2026-09-10
dev has not had master merged back into it since v2.3.1 — release tags v2.4.0-v2.6.1 are invisible from dev


webbpinner:
## Summary

`dev` and `master` have diverged: `master` has cut five releases (`v2.4.0`, `v2.5.0`, `v2.5.1`, `v2.6.0`, `v2.6.1`) since March 2026 that were never merged back into `dev`. `dev`'s git history hasn't had `master` merged into it since `v2.3.1` (2026-03-03). This doesn't lose any code — `dev`'s own content is current (and ahead, via its own unreleased commits) — but it does mean any tooling that walks git tag ancestry from `dev` (e.g. `git describe`, `setuptools_scm`) reports a version far behind reality.

## How this was discovered

While testing #618 (display the OpenRVDAS version in the UI, switched to deriving the version from git tags via `setuptools_scm`), a local install from the `dev`-based branch reported `2.3.2.dev82+g...` instead of the expected `2.6.1.dev<something>`. Investigating why surfaced this:

```
$ git merge-base --is-ancestor v2.6.1 origin/dev
exit 1   # v2.6.1 is not an ancestor of dev

$ git describe --tags origin/dev
v2.3.1-79-gfdf6952e   # dev's nearest reachable tag is v2.3.1, 79 commits back

$ git rev-list --count origin/dev..origin/master
1        # master has only 1 commit dev doesn't have (the v2.6.1 cut)

$ git rev-list --count origin/master..origin/dev
116      # dev has 116 commits of its own unreleased work since the last shared point

$ git log -1 --format="%H %ai %s" v2.3.1
ba960750... 2026-03-03 16:46:45 -0800 Merge branch 'master' into dev

$ git log -1 --format="%H %ai %s" v2.6.0
15fea5d3... 2026-08-16 19:19:55 -0700 Release v2.6.0: merge dev → master (#587)
```

## Root cause

Releases are cut by merging `dev` into `master` and tagging the merge commit (e.g. `v2.6.0`'s message: "Release v2.6.0: merge dev → master"). At some point the *reverse* step — merging `master` back into `dev` and tagging that merge (which is what produced the `v2.3.1` tag, itself a "Merge branch 'master' into dev" commit) — stopped happening. Every release since `v2.3.1` (`v2.4.0` through `v2.6.1`) only exists in `master`'s history, invisible to anything walking `dev`'s ancestry, including `git describe`/`setuptools_scm`-based versioning.

## Impact

[...]

Issue #620, opened 2026-09-10
install_openrvdas.sh: refuse to install the React UI when python3 is below the 3.11 floor web_backend requires


webbpinner:
## Summary

`utils/install_openrvdas.sh`'s `setup_new_ui_backend` (called when `UI_TYPE=react`) builds the `web_backend` virtualenv using bare `python3` on every non-macOS target:

```bash
if [ "${OS_TYPE:-}" == 'MacOS' ]; then
    _BREW_PFX="..."
    BACKEND_PYTHON="${_BREW_PFX}/opt/python@3.12/bin/python3.12"
else
    BACKEND_PYTHON=python3
fi
```

`web_backend/pyproject.toml` requires `python = "^3.11"`, and `web_backend/app/main.py` does `import tomllib`, which is stdlib-only since Python 3.11. There is currently no check that bare `python3` actually satisfies that floor before the venv is built, so on any target where the system's default `python3` is older than 3.11, the install completes "successfully" and the FastAPI backend then fails at startup with an import error — a confusing failure mode discovered while working on #618 and documented in the audit in #619.

Per the #619 audit, this currently bites even actively-supported targets: Ubuntu 22.04 (`python3.10`) and RHEL/Rocky/Alma/CentOS 8 & 9 (`python3.8`/`python3.9`).

## Proposed fix

Before (or inside) `setup_new_ui_backend`, when `UI_TYPE=react`, verify that the Python that will be used to build the `web_backend` venv is >= 3.11, and fail the install with a clear, actionable error message if not — rather than letting it silently proceed and break at runtime. For example:

```bash
if [[ "$UI_TYPE" == "react" ]] && [ "${OS_TYPE:-}" != 'MacOS' ]; then
    if ! python3 -c 'import sys; sys.exit(0 if sys.version_info >= (3, 11) else 1)'; then
        echo "ERROR: The React/FastAPI UI requires Python 3.11+, but this system's"
        echo "python3 is $(python3 --version). Install a Python 3.11+ interpreter"
        echo "(e.g. via deadsnakes PPA on Ubuntu, or the EPEL python3.11 package on"
        echo "RHEL/Rocky/Alma) and re-run, or choose UI_TYPE=django/none instead."
        exit_gracefully
    fi
fi
```

[...]

Issue #619, opened 2026-09-10
Decide which OS/versions to keep supporting in install_openrvdas.sh, given EOL status and the web_backend Python 3.11 requirement


webbpinner:
## Summary

`utils/install_openrvdas.sh` (`get_os_type`) currently accepts a fairly wide range of Ubuntu, Debian, Linux Mint, and RHEL-family OS versions. Several of these are now past their vendor end-of-life date, and — separately — the Python version each one resolves to for `python3` isn't tracked against what OpenRVDAS itself needs (`>=3.8` for the core package; `^3.11` for the `web_backend` FastAPI submodule, since `web_backend/app/main.py` uses the stdlib `tomllib`, which requires Python 3.11+).

This issue is to **decide which of these OS/version combinations we continue to officially support**, using their EOL status and Python-version story as inputs, and to update `get_os_type`'s accepted list (and README/docs) accordingly.

## Data: every OS/version the script currently accepts (as of 2026-09-10)

| `OS_TYPE` | `OS_VERSION` | Real release | Default `python3` | EOL date | Currently supported by vendor? | `python3.11`+ installable? |
|---|---|---|---|---|---|---|
| Ubuntu | 18 | 18.04 Bionic | 3.6.9 | Standard support ended 2023-05-31 (ESM to 2028) | No | No — deadsnakes PPA no longer builds for bionic |
| Ubuntu | 20 | 20.04 Focal | 3.8.10 | Standard support ended 2025-05-31 (ESM to 2030) | No | No — deadsnakes PPA no longer builds for focal |
| Ubuntu | 21 | 21.04/21.10 (interim) | 3.9 | EOL ~2022 | No | N/A — long dead |
| Ubuntu | 22 | 22.04 Jammy | 3.10.x | 2027-05-31 | **Yes** | **Yes** — deadsnakes PPA supports jammy |
| Ubuntu | 23 | 23.04/23.10 (interim) | 3.11 | EOL ~2024 | No | N/A — long dead |
| Ubuntu | 24 | 24.04 Noble | 3.12.x | 2029-05-31 | **Yes** | Already ≥3.11 natively |
| Ubuntu | 25 | 25.04/25.10 (interim) | 3.13 | 25.04 EOL 2026-01, 25.10 EOL 2026-07 | No | N/A — both already EOL |
[...]

PR #622, opened 2026-09-10
[Issue #618] Display the OpenRVDAS version in both the React and Django UIs

### LINKED ISSUE #618: Display current version number in the UI
## Summary

Display the current OpenRVDAS version number in the UI, in both the React frontend and the Django GUI.

## Background

- The React frontend (`web_frontend/`) renders a version in `TopNav.tsx` and the document title (`web_frontend/src/App.tsx:68`), sourced from `AppConfig.version` / the `VITE_OPENRVDAS_VERSION` env var (`web_frontend/src/config.ts:13`).
- `utils/install_openrvdas.sh:1253-1258` sets `VITE_OPENRVDAS_VERSION` in `web_frontend/.env` at **install** time by reading `pyproject.toml`. This only runs at install, so if `pyproject.toml`'s version is later bumped (e.g. `git pull`/upgrade of an already-installed system) without re-running the installer, the displayed version goes stale. It also falls back to `.env.dist`'s hardcoded `dev` or `web_frontend/package.json`'s unrelated `version` field (`1.0.0`) if unset.
- `web_backend/app/main.py:20,31` already reads `project_version` from `pyproject.toml` (via poetry) and uses it for the FastAPI OpenAPI `version` field.
- The Django GUI (`django_gui/`), the primary management console per `CLAUDE.md`, has no version display anywhere in its templates or views.
- The React UI is never deployed without `web_backend` (FastAPI) running, so the frontend can safely depend on it for version info at runtime.

## Version source: switched to setuptools_scm (git tags), not a static pyproject.toml field

While implementing this, we found `pyproject.toml`'s `version = "0.1.0"` had drifted badly from reality — the repo's actual releases (via git tags/GitHub Releases) are up to `v2.6.1`, but nothing was ever updating the hand-maintained field. Rather than add a manual/CI bump step to keep a static string in sync (more process to maintain, still driftable), `pyproject.toml` now declares `dynamic = ["version"]` and uses [`setuptools_scm`](https://setuptools-scm.readthedocs.io/) to derive the version directly from git tags at install time — the tag *is* the source of truth, so there's nothing to keep in sync. A tagged commit gets a clean version (`2.6.1`); an untagged commit gets a PEP 440 dev version (e.g. `2.6.2.dev3+g1234567`).

This changes how the version is read at runtime, everywhere:
- `logger/utils/read_version.py` (used by the Django context processor) now calls `importlib.metadata.version("openrvdas")` instead of parsing `pyproject.toml` with a regex, falling back to `"unknown"` if OpenRVDAS isn't installed as a package in the current environment.
- `web_backend/app/main.py`'s `project_version` now also comes from `importlib.metadata.version("openrvdas")` instead of reading either its own or the parent's `pyproject.toml` via `tomllib`.
- Since `importlib.metadata.version()` only works once a package is actually `pip install`-ed (not just present on `sys.path`, which is how `web_backend` normally reaches OpenRVDAS's modules), `utils/install_openrvdas.sh`'s `setup_new_ui_backend` now also runs `pip install --no-deps -e "${INSTALL_ROOT}/openrvdas"` into the backend's venv, purely to register OpenRVDAS's package metadata there (`--no-deps` avoids pulling OpenRVDAS's full dependency set into that venv, since the backend still reaches the actual modules via `sys.path`, not this install).
- The main OpenRVDAS venv already runs `pip install -e .`, so it picks up the setuptools_scm-derived version with no additional change needed there.

One caveat worth knowing: `dev` and `master` have diverged tag histories (recent release tags like `v2.4.0`–`v2.6.1` exist only in `master`'s ancestry), so a `dev` checkout won't show `2.6.1` — it'll show a dev-version relative to the last tag reachable in `dev`'s own history. That's expected/correct setuptools_scm behavior, not a bug, but it does mean the two branches will visibly disagree on version number until they're reconciled.

## Proposed solution

- [x] Add a small `GET /api/v1/version` endpoint to `web_backend` returning the installed version (via `importlib.metadata.version("openrvdas")`).
- [x] Update the React frontend to fetch the version from that endpoint at runtime instead of reading `AppConfig.version` from `VITE_OPENRVDAS_VERSION`/`package.json`.
- [x] Remove the now-unneeded `VITE_OPENRVDAS_VERSION` env var, its `.env`/`.env.dist` entries, and the install-time sync step in `utils/install_openrvdas.sh:1253-1258`.
- [x] Add a Django context processor that reads the version and injects it into template context; display it in `django_gui/templates/django_gui/base.html`'s footer.
- [x] Switch `pyproject.toml` to `setuptools_scm`-derived versioning instead of a hand-maintained static string.

This removes the staleness problem entirely — both for the old install-time `.env` sync (no sync step to forget or re-run on upgrade) and for the version number itself (no hand-bumped field to forget to update on release).

## Why

There's currently no version display in the Django GUI at all, the React frontend's version could go stale on upgraded (not freshly reinstalled) systems, and separately, `pyproject.toml`'s version field had already drifted far from the repo's actual release history.

webbpinner:
## Summary
- **Django GUI**: adds a `openrvdas_version` context processor (`django_gui/context_processors.py`), registered in `settings.py.dist`, displayed in `django_gui/templates/django_gui/base.html`'s footer.
- **`pyproject.toml`**: switched from a hand-maintained `version = "0.1.0"` field (badly drifted from the repo's real releases, up to `v2.6.1`) to `setuptools_scm`-derived versioning from git tags — no field to remember to bump.
- **`logger/utils/read_version.py`**: reads the installed package's version via `importlib.metadata.version("openrvdas")`.
- **`utils/install_openrvdas.sh`**: drops the now-unneeded `VITE_OPENRVDAS_VERSION` install-time sync step, and adds `pip install --no-deps -e "${INSTALL_ROOT}/openrvdas"` into the `web_backend` venv in `setup_new_ui_backend`, so `importlib.metadata` can find OpenRVDAS's metadata there too (for the FastAPI backend's own version endpoint).
- Bumps the `web_backend` and `web_frontend` submodule pointers.

## Related
Closes #618. Should merge together with (or after) OceanDataTools/fastapi-template#1 and OceanDataTools/vite-react-template#1 — the submodule pointers here reference commits on those PRs' branches.

Also surfaced two follow-up issues during this work, filed separately (not addressed here): #619 (OS/version support audit) → #620 (install script should refuse the React UI when Python < 3.11) → #621 (`dev` hasn't had `master` merged back into it since `v2.3.1`, which is why a `dev`-based install currently shows a version like `2.3.2.dev82+g...` instead of something near `2.6.1`).

## Test plan
- [x] `pytest test/logger/utils/` — 133/133 pass
- [x] `flake8` on all changed Python files — clean
- [x] `bash -n utils/install_openrvdas.sh` — syntax OK
- [x] Manually verified end-to-end locally: ran the FastAPI backend, React dev server, and Django dev server together; all three independently reported the same `setuptools_scm`-derived version string

https://claude.ai/code/session_015eqosA985SzDWzzEoFqLMH


webbpinner:
Screenshot 2026-09-10 at 10-47-47 OpenRVDAS Cruise Management

webbpinner:
sub-module PR are already merged.

pyproject.toml
```diff
@@ -128,6 +128,12 @@ openrvdas_manage = "openrvdas_cli:openrvdas_manage"
 py-modules = ["openrvdas_cli", "manage"]
 include-package-data = true
 
+# Version is derived from git tags (e.g. v2.6.1) rather than hand-maintained
+# here, so it can never drift from what's actually released. Untagged
+# commits get a PEP 440 dev version like "2.6.2.dev3+g1234567".
+[tool.setuptools_scm]
```
davidpablocohn, line 143: **No `fallback_version` — builds now fail outside a git tree.**

`setuptools_scm` raises `LookupError: setuptools-scm was unable to detect version` when `.git` is absent. The old `version = "0.1.0"` always succeeded, so this is a behavior regression for:

- GitHub's "Download ZIP"
- an rsync'd or `cp -r`'d source tree
- `docker COPY . .` with `.git` in `.dockerignore`

One line restores the old behavior with an obviously-bogus marker:

```toml
fallback_version = "0.0.0"
```

pyproject.toml
```diff
@@ -128,6 +128,12 @@ openrvdas_manage = "openrvdas_cli:openrvdas_manage"
...
+tag_regex = "^v(?P<version>\\d+\\.\\d+\\.\\d+.*)$"
```
davidpablocohn: **This regex rejects 9 of the repo's own tags.**

`^v(?P<version>\d+\.\d+\.\d+.*)$` doesn't match the legacy tags still in the repo:

```
2.0.0  a2.0.0  a2.0.1  a2.0.2  a2.0.3  a2.0.4  a2.0.5  b1.9.1  b1.9.2
```

If `git describe --tags --match '*[0-9]*'` ever selects one — a build from a branch whose nearest reachable matching tag is legacy — setuptools_scm's `meta()` trips `assert parsed_version is not None` and the build dies with a bare `AssertionError` rather than falling back.

setuptools_scm's stock default `tag_regex` accepts both forms, and the custom regex buys nothing here since `packaging.Version` already normalizes the leading `v`. Suggest just dropping the override.

**Related edge, same line:** `dev` currently carries a `v2.6.2.dev0` marker tag (added while fixing #621). setuptools_scm's `_bump_dev` asserts `tail == "0"` on a `.dev`-suffixed tag, so `v2.6.2.dev0` is fine but a future `v2.6.2.dev1` would fail the build with `'own dev numbers are unsupported'`. Worth knowing that the marker convention is limited to one `.dev0` per release cycle — I'll note it in `RELEASING.md`.

pyproject.toml
```diff
@@ -1,10 +1,10 @@
 [build-system]
-requires = ["setuptools>=68", "wheel"]
+requires = ["setuptools>=68", "setuptools_scm>=8", "wheel"]
```
davidpablocohn: **Unbounded `setuptools_scm>=8`.**

Build requirements resolve fresh on every install (no lock), and 9.x and 10.x have both shipped since 8. This is the code path that decides whether `pip install -e .` succeeds at all, so a breaking change in a future major silently breaks every new install.

Suggest `"setuptools_scm>=8,<11"`, consistent with the version-ceiling policy documented further down this file.

django_gui/context_processors.py
```diff
@@ -0,0 +1,6 @@
+from logger.utils.read_version import get_version
+
+
+def openrvdas_version(request):
+    """Make the OpenRVDAS release version available in every template context."""
+    return {'openrvdas_version': get_version()}
```
davidpablocohn: **Runs on every page render.**

Context processors execute per template render, so this calls `importlib.metadata.version()` on every Django request. That's uncached — it re-walks each `sys.path` entry's `*.dist-info` each time.

Worse, in an environment where the package isn't pip-installed, `read_version.py:28`'s `logging.warning(...)` fires once per request and spams the log.

The version can't change within a process lifetime, so compute it once at import:

```python
_VERSION = get_version()

def openrvdas_version(request):
    return {'openrvdas_version': _VERSION}
```

django_gui/templates/django_gui/base.html
```diff
@@ -11,7 +11,7 @@
     </div>
     <footer>
       <a href="https://github.com/oceandatatools/openrvdas/blob/master/README.md" target="_blank">
-        OpenRVDAS &copy;<script>document.write(new Date().getFullYear())</script> David Pablo Cohn
+        OpenRVDAS v{{ openrvdas_version }} &copy;<script>document.write(new Date().getFullYear())</script> David Pablo Cohn
```
davidpablocohn: **Two degraded renders on this line.**

1. When the package isn't pip-installed, `get_version()` returns `'unknown'` and the footer reads `OpenRVDAS vunknown`.
2. `django_gui/settings.py` is gitignored and regenerated only by `setup_django`. Any deployment upgraded without re-running the installer keeps a `settings.py` with no `openrvdas_version` context processor registered — Django resolves the unknown variable to `''`, and the footer renders a dangling `OpenRVDAS v ©2026`.

Guarding covers both:

```html
{% if openrvdas_version %}OpenRVDAS v{{ openrvdas_version }} {% endif %}&copy;...
```

logger/utils/read_version.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
+"""Utility for reading the installed OpenRVDAS release version.
```
davidpablocohn: **The "can't drift" claim doesn't hold for editable installs.**

For the editable install this project actually uses, the version string is frozen into dist-info `METADATA` at `pip install` time — it is not recomputed from git on import.

So any upgrade that doesn't re-run `pip install -e .` — a plain `git pull` in `/opt/openrvdas`, which is a common field workflow — leaves all three UIs reporting the old version indefinitely, with nothing indicating it's stale. That's arguably worse than the drifting `pyproject.toml` field it replaces, because it looks authoritative.

Worth softening the docstring and noting that a reinstall is required for the displayed version to update.

utils/install_openrvdas.sh
```diff
@@ -1202,6 +1202,15 @@ ENVDEFAULTS
         --trusted-host pypi.org --trusted-host files.pythonhosted.org \
         -r "$BACKEND_DIR/requirements.txt"
 
+    # Register OpenRVDAS's own package metadata in this venv (--no-deps: the
+    # backend reaches OpenRVDAS's modules via sys.path, not this install; this
+    # is only so importlib.metadata.version("openrvdas") can find it, for
+    # GET /api/v1/version). Version comes from git tags via setuptools_scm.
+    echo "Registering OpenRVDAS package metadata in web backend venv..."
```
davidpablocohn, line 1209: **The comment understates what this does.**

`pip install --no-deps -e` doesn't just register metadata — it also installs all 16 console-script entry points (`listen`, `logger_manager`, `cached_data_server`, `logger_runner`, ...) into `$BACKEND_VENV/bin/`, and puts `logger`, `server`, `database`, `utils`, `django_gui`, `manage` and `openrvdas_cli` on the backend venv's import path — all *without* their runtime dependencies, since `--no-deps` skips them.

Anything resolving those scripts from the backend venv's PATH gets an `ImportError` rather than a clean "not found". (I checked `web_backend` for top-level module-name collisions against those packages — none today, so imports are safe.)

Not necessarily wrong, but the comment should say what's actually happening.

davidpablocohn (commented):
Read through all six files plus both submodule pointer diffs, and ran the new tests and flake8 on a worktree of the branch — 3/3 pass, flake8 clean.

**No blocking correctness defect.** The wiring checks out: `apiUrl("/version")` resolves to the backend's new `/api/v1/version`, no dangling `VITE_OPENRVDAS_VERSION` references remain in either repo, and the install ordering (`git checkout` -> `pip install -e .` -> `setup_new_ui_backend`) is correct. Comments below are robustness and edge cases, left inline.

Two things that can't be commented inline, since they're inside the submodules:

- **`web_backend` Dockerfile** — only `setup_new_ui_backend` in the install script registers OpenRVDAS metadata in the backend venv. The backend's own Dockerfile does `poetry install --only main --no-root` and never installs OpenRVDAS, so `/api/v1/version` returns `{"version": "unknown"}` in every Docker/compose deployment, and the version disappears from the React nav and page title there.
- **`web_frontend/src/hooks/useAppVersion.ts`** — the hook is mounted independently in both `App` and `TopNav`, so each page load issues two identical `GET /api/v1/version` requests, and it refetches on every remount. A module-level cached promise or a Redux slice would make it one per session.

And a sequencing note: `RELEASING.md` (merged yesterday in #625) has a "Version numbers" section that describes `setuptools_scm` versioning as *in progress* and tells maintainers to hand-bump `pyproject.toml`'s `version` field — the field this PR removes. That section needs updating in this PR or immediately after it.

webbpinner:
Figures that this was what I originally dug in to the code for... and it's the only piece with issues...  thanks for double checking the work.  Will revise.

pyproject.toml
davidpablocohn: Note: I think it's fine to not be able to not match the legacy tags - I can't imagine anyone still using any of those versions.

davidpablocohn: Agreed on the legacy tags — happy to drop that half. One clarification just so the call is made on the right risk: the concern wasn't that anyone still runs `v1.9`, it's that `git describe` could *select* one of those tags during a build (only on a branch with no modern tag reachable) and crash the build rather than fall back. That's a narrow enough case that ignoring it seems reasonable; just flagging the mechanism in case it read as a compatibility argument.

Separately, **I need to correct the `.dev` note at the bottom of my comment** — I had the specifics wrong. It is not an `AssertionError` from `_bump_dev`, and the message isn't `'own dev numbers are unsupported'`. Tested against setuptools_scm 9.x:

```
ValueError: choosing custom numbers for the `.devX` distance is not supported.
 The 2.6.2.dev1 can't be bumped
Please drop the tag or create a new supported one ending in .dev0
```

The more useful detail I also got wrong: it does **not** fail when the tag is created. At distance 0 a `.dev1` tag resolves fine — it only raises once the next commit lands, by which point the bad tag may already be pushed.

| tag | at tag | + 1 commit |
|---|---|---|
| `v2.9.9.dev0` | `2.9.9.dev0` | `2.9.9.dev1+g<sha>` |
| `v2.9.9.dev1` | `2.9.9.dev1` | **`ValueError`** |
| `v2.9.9` | `2.9.9` | `2.9.10.dev1+g<sha>` |

So the constraint stands — marker tags must end in `.dev0`, and setuptools_scm derives the running count itself — but nothing in this PR needs to change for it. I've written it up in #626 against `RELEASING.md` instead. Sorry for the noise.

webbpinner:
@davidpablocohn Thanks for the thorough review — pushed fixes for everything raised:

1. **`fallback_version` missing** → added `fallback_version = "0.0.0"` under `[tool.setuptools_scm]`, so builds with no `.git` tree (zip download, rsync'd tree, `docker COPY` excluding `.git`) still succeed instead of raising. (cc0ef4b7)
2. **Custom `tag_regex` rejects legacy tags** → dropped the override entirely; confirmed locally that stock setuptools_scm's default already handles the `v` prefix and resolves the same version string. (116a5c7e)
3. **Unbounded `setuptools_scm>=8`** → capped at `>=8,<11`, consistent with this file's existing version-ceiling convention. (cc0ef4b7)
4. **Context processor recomputing on every request** → version is now computed once at module import (`_VERSION = get_version()`) instead of per-render. (cc0ef4b7)
5. **Degraded footer renders** (`vunknown`, dangling `v ©`) → footer now guards with `{% if openrvdas_version and openrvdas_version != 'unknown' %}`, covering both the uninstalled-package case and the stale-`settings.py`-without-the-context-processor case. (cc0ef4b7)
6. **"Can't drift" claim not true for editable installs** → `get_version()` now asks `setuptools_scm` to recompute live from the current git tree on every call (still cached once per process by the context processor), falling back to the frozen install-time metadata only when there's no `.git` tree or `setuptools_scm` isn't available. Added `setuptools_scm` as a runtime dependency (was build-only) so it's actually importable post-install, and updated/expanded the tests to cover the fallback chain. (5bd8c199)

Left open / out of scope for this PR:
- The `install_openrvdas.sh` comment on the `--no-deps -e` backend install understating what it registers — still accurate as a smaller wording fix, happy to take it in a follow-up if you'd like it here instead.
- The `web_backend` Dockerfile not installing OpenRVDAS (so `/api/v1/version` returns `"unknown"` in Docker/compose) and `useAppVersion.ts` double-fetching — both live in the submodules, so tracking separately rather than in this PR.

All new/changed tests pass (`pytest test/logger/utils/test_read_version.py` — 5/5) and flake8 is clean on the touched files.

webbpinner:
@davidpablocohn All six review points are addressed (see the summary above) and CI is green. Could you take another look when you get a chance? Also filed follow-ups for the two submodule findings you couldn't comment on inline: OceanDataTools/fastapi-template#2 and OceanDataTools/vite-react-template#2 — treating those as non-blocking for this PR unless you feel otherwise.

davidpablocohn (approved):

